PR #50, opened 2016-04-02
Fall back to non-PSK if no shared PSK ciphers

grittygrease:


yaronf:
Hi Nick, I don't see a use case where this would make sense. If client and server have a PSK, they surely want mutual authentication even if the server happens to have a cert.

grittygrease:
The use case is if the client is sending a valid session ticket in a resumption but the client and server don't share the same PSK ciphersuites: then you have to fall back to a regular non-resumption handshake.

yaronf:
Agree. Thanks.

PR #55, opened 2016-04-03
Document workaround for running tests

kaduk:
In some cases, some people have reported that the tests fail to run.
Manually setting DIST and OBJDIR seems to be an effective workaround.


bifurcation:
LGTM

bifurcation:
@martinthomson Is this needed after your refactor in #63?

martinthomson:
Not really needed any more.  You just don't get to choose where the objects end up, which is annoying later when you use them, but not fatal.

PR #58, opened 2016-04-03
Fix 0RTT so server doesn't wait for end of client data

ekr:


conn.go
```diff
@@ -1065,6 +986,87 @@ func (c *Conn) serverHandshake() error {
 		return err
 	}
 
+
```
bifurcation: Extra newline

PR #61, opened 2016-04-05
Fix buffer issue by limiting max plaintext size.

grittygrease:


bifurcation:
This PR is long obsolete.

PR #66, opened 2016-05-09
Support DHE by default, add DHE_PSK

### LINKED ISSUE #64: Support DHE by default
No reason not to include this, well, except for the fact that the ClientHello will bloat.

### LINKED ISSUE #65: Support DHE-based resumption
TLS_DHE_PSK_WITH_AES_128_GCM_SHA256 and TLS_DHE_PSK_WITH_AES_256_GCM_SHA384

martinthomson:
Closes #65 
Closes #64


martinthomson:
Note that I managed to get interop with my implementation in NSS, but only after adding c709712.

conn.go
```diff
@@ -174,13 +174,19 @@ var (
...
 		TLS_ECDHE_PSK_WITH_AES_128_GCM_SHA256,
+		TLS_ECDHE_PSK_WITH_AES_256_GCM_SHA384,
+		TLS_DHE_PSK_WITH_AES_128_GCM_SHA256,
+		TLS_DHE_PSK_WITH_AES_256_GCM_SHA384,
 	}
 
 	defaultSupportedGroups = []namedGroup{
 		namedGroupP256,
 		namedGroupP384,
-		namedGroupP521,
```
bifurcation: Why are you removing P-521?

bifurcation: Noting for the historical record that this patch is removing P-521 from the default list of groups.  It's unrelated, but I'll let it slide.

PR #70, opened 2016-06-09
Use crypto.Hash as crypto.SignerOpts

FiloSottile:


bifurcation:
This has been merged, but for some reason, GitHub isn't picking it up.

PR #72, opened 2016-07-12
Support for draft-13

bifurcation:
There may still be some bugs in here, but this at least does a basic handshake with NSS, both as client and server (with NSS in complementary roles).

### DIFF
--- a/handshake-messages.go
+++ b/handshake-messages.go
@@ -7,14 +7,17 @@ import (
 )
 
 const (
-	fixedClientHelloBodyLen  = 39
-	fixedServerHelloBodyLen  = 36
-	maxCipherSuites          = 1 << 15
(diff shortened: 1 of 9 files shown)


handshake-messages.go
```diff
@@ -440,42 +443,65 @@ func (cv *certificateVerifyBody) computeContext(transcript []*handshakeMessage)
...
+//
+// struct {
+//     TicketExtensionType extension_type;
+//     opaque extension_data<1..2^16-1>;
+// } TicketExtension;
+//
+// struct {
+//     uint32 ticket_lifetime;
+//     uint32 flags;
+//     uint32 ticket_age_add;
```
martinthomson: Watch out: looks like you have -14 comments.

PR #74, opened 2016-07-16
use the right environment variable to configure NSS

ekr:


bifurcation:
This appears to no longer be necessary.

PR #80, opened 2016-08-03
Add support for ALPN

### LINKED ISSUE #21: Support NPN / ALPN

bifurcation:
Fixes #21 

r? @ekr

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -71,6 +71,7 @@ type Config struct {
 	// Hidden fields (used for caching in convenient form)
 	enabledSuite map[cipherSuite]bool
 	enabledGroup map[namedGroup]bool
+	enabledProto map[string]bool
 	certsByName  map[string]*Certificate
 
 	// The same config object can be shared among different connections, so it
@@ -127,6 +128,7 @@ func (c *Config) Init(isClient bool) error {
 	// Build caches
 	c.enabledSuite = map[cipherSuite]bool{}
 	c.enabledGroup = map[namedGroup]bool{}
+	c.enabledProto = map[string]bool{}
 	c.certsByName = map[string]*Certificate{}
 
 	for _, group := range c.Groups {
@@ -153,6 +155,9 @@ func (c *Config) Init(isClient bool) error {
 				c.enabledSuite[suite] = true
 			}
 		}
+	}
+	for _, proto := range c.NextProtos {
+		c.enabledProto[proto] = true
 	}
 	logf(logTypeCrypto, "Enabled suites [%v]", c.enabledSuite)
 
@@ -524,6 +529,11 @@ func (c *Conn) clientHandshake() error {
 	sa := signatureAlgorithmsExtension{algorithms: c.config.SignatureAlgorithms}
 	dv := draftVersionExtension{version: draftVersionImplemented}
 
+	var alpn *alpnExtension
+	if (c.config.NextProtos != nil) && (len(c.config.NextProtos) > 0) {
+		alpn = &alpnExtension{protocols: c.config.NextProtos}
+	}
+
 	var psk *preSharedKeyExtension
 	if key, ok := c.config.ClientPSKs[c.config.ServerName]; ok {
 		logf(logTypeHandshake, "Sending PSK")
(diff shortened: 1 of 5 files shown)


conn.go
```diff
@@ -66,10 +66,12 @@ type Config struct {
 	CipherSuites        []cipherSuite
 	Groups              []namedGroup
 	SignatureAlgorithms []signatureAndHashAlgorithm
+	NextProtos          []string
```
ekr: I feel like this should be an array of bytes instead.

conn.go
```diff
@@ -523,6 +529,11 @@ func (c *Conn) clientHandshake() error {
 	sa := signatureAlgorithmsExtension{algorithms: c.config.SignatureAlgorithms}
 	dv := draftVersionExtension{version: draftVersionImplemented}
 
+	var alpn *alpnExtension
+	if (c.config.NextProtos != nil) && (len(c.config.NextProtos) > 0) {
```
ekr: Can this kind of vector be nil?

conn.go
```diff
@@ -851,6 +873,19 @@ func (c *Conn) serverHandshake() error {
 		}
 	}
 
+	// Find early_data extension and handle early data
```
ekr: This comment seems wrong

conn.go
```diff
@@ -851,6 +873,19 @@ func (c *Conn) serverHandshake() error {
...
+	sendALPN := false
+	var serverALPN *alpnExtension
+	if gotALPN {
+		for _, proto := range clientALPN.protocols {
+			if c.config.enabledProto[proto] {
```
ekr: I would tend to prefer searching through the list here rather than building an unnecessary map.

conn.go
```diff
@@ -1061,6 +1096,13 @@ func (c *Conn) serverHandshake() error {
 			return err
 		}
 	}
+	if sendALPN {
```
ekr: Why can't you just add the extension above or at least use serverALPN as the conditional

extensions.go
```diff
@@ -567,6 +567,56 @@ func (ed *earlyDataExtension) Unmarshal(data []byte) (int, error) {
...
+
+func (alpn alpnExtension) Type() helloExtensionType {
+	return extensionTypeALPN
+}
+
+func (alpn alpnExtension) Marshal() ([]byte, error) {
+	listData := []byte{}
+	for _, proto := range alpn.protocols {
+		listData = append(listData, byte(len(proto)))
+		listData = append(listData, []byte(proto)...)
```
ekr: I don't know go well, what does the ... do?

extensions.go
```diff
@@ -567,6 +567,56 @@ func (ed *earlyDataExtension) Unmarshal(data []byte) (int, error) {
...
+	}
+
+	listLen := len(listData)
+	lenData := []byte{byte(listLen >> 8), byte(listLen)}
+	return append(lenData, listData...), nil
```
ekr: You can't append integers directly?

extensions.go
```diff
@@ -567,6 +567,56 @@ func (ed *earlyDataExtension) Unmarshal(data []byte) (int, error) {
...
+	if len(data) < 2+listLen {
+		return 0, fmt.Errorf("tls.alpn: Too short for proto list")
+	}
+
+	read := 2
+	alpn.protocols = []string{}
+	for read < listLen+2 {
+		itemLen := int(data[read])
+		if 2+listLen < read+itemLen {
+			return 0, fmt.Errorf("tls.alpn: Too short for list")
```
ekr: Isn't this too long?

extensions.go
```diff
@@ -567,6 +567,56 @@ func (ed *earlyDataExtension) Unmarshal(data []byte) (int, error) {
...
+	listLen := (int(data[0]) << 8) + int(data[1])
+	if len(data) < 2+listLen {
+		return 0, fmt.Errorf("tls.alpn: Too short for proto list")
+	}
+
+	read := 2
+	alpn.protocols = []string{}
+	for read < listLen+2 {
+		itemLen := int(data[read])
+		if 2+listLen < read+itemLen {
```
ekr: Why not increment read here?

ekr:
This looks basically right module the string/byte array issues noted above. I notice you have no functional tests for alpn match/mismatch

conn.go
bifurcation: `crypto/tls` has it as `[]string`, so I'm going to keep it to be consistent.

conn.go
bifurcation: Yep.  Technically, it's a "slice" vs. an "array", so it can be nil.

extensions.go
bifurcation: That's how you append another slice, vs. appending a single element (as on the previous line).

PR #84, opened 2016-08-09
Fix PSK compatibility issues

### LINKED ISSUE #77: Pre-Shared Key extension incorrect in ServerHello
I have a question regarding the encoding of the PSK identity in the ServerHello in the Mint implementation. 

I received the following Pre-Shared Key extension in a ServerHello:

> 00 09 00 29 00 05 00 03 61 62 63
 ...)....abc

The PSK identity in this case is ‘abc’ and it is 3 bytes long. 
The Pre-Shared Key extension has the value 41 (or 0x29). 

The format is

      

    struct {
       ExtensionType extension_type;
       opaque extension_data<0..2^16-1>;
      } Extension;

     opaque psk_identity<0..2^16-1>;

      struct {
           
       select (Role) {
           case client:
               psk_identity identities<2..2^16-1>;

           case server:
               uint16 selected_identity;
       }
      } PreSharedKeyExtension;

I would have expected to see 

> 00 29 00 02 00 00

In case that the selected identity from the ClientHello was the first one. 

The format of the Pre-Shared Key extension encoding has changed recently but wouldn’t the encoding for the old format still be something like; 

> 00 29 00 05 00 03 61 62 63
 ...)....abc

### LINKED ISSUE #78: Check for missing extensions incorrect
The code in conn.go incorrectly requires various extensions to be present. This causes problems when using a PSK-based ciphersuite since neither supported groups nor signature algorithm extensions need to be present (or are even useful). Even the SNI extension is not necessary. 

This check should be much later in the code when it is clear what ciphersuite will be selected. 

The relevant code is here: 

> if !gotServerName || !gotSupportedGroups || !gotSignatureAlgorithms {
		logf(logTypeHandshake, "[server] Insufficient extensions")
		return fmt.Errorf("tls.server: Missing extension in ClientHello (%v %v %v %v)",
			gotServerName, gotSupportedGroups, gotSignatureAlgorithms, gotKeyShares)
	}

bifurcation:
Fixes #77 
Fixes #78

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -665,7 +665,7 @@ func (c *Conn) clientHandshake() error {
 	foundKeyShare := sh.extensions.Find(&serverKeyShare)
 
 	var pskSecret, pskContext, dhSecret []byte
-	if foundPSK && psk.HasIdentity(serverPSK.identities[0]) {
+	if foundPSK && (serverPSK.selectedIdentity < uint16(len(psk.identities))) {
 		pskSecret = c.config.ClientPSKs[c.config.ServerName].Key
 		pskContext = c.config.ClientPSKs[c.config.ServerName].Context
 		logf(logTypeHandshake, "[client] got PSK extension")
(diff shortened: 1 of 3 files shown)


bifurcation:
@hannestschofenig r?

PR #98, opened 2016-11-10
NSS Interop

bifurcation:


crypto.go
```diff
@@ -802,15 +802,24 @@ func (ctx *cryptoContext) preInit(psk PreSharedKey) error {
 	if psk.IsResumption {
 		binderLabel = labelResumptionBinder
 	}
-	ctx.binderKey = ctx.deriveSecret(ctx.earlySecret, binderLabel, []byte{})
+
+	// XXX: Spec says "", but NSS uses []byte{0}.
+	// TODO: File a spec bug
+	h := ctx.params.hash.New()
+	h.Write([]byte{0})
```
franziskuskiefer: This has to be h.Write([]byte{}).

It's still not working, but this gives us at least the correct binder key.

PR #99, opened 2016-11-12
Have the server send EDI when accepting 0-RTT. 0-RTT interop with NSS…

ekr:
… works now


bifurcation:
Merging while red because tests are busted on nss-interop right now.  Will fix the tests once we get ready to merge nss-interop.

PR #115, opened 2016-11-30
Implement obfuscated ticket age on the client - closes #114

### LINKED ISSUE #114: Implement obfuscated_ticket_age

FiloSottile:


conn_test.go
```diff
@@ -371,19 +371,6 @@ func TestResumption(t *testing.T) {
...
-
-	var serverPSK PreSharedKey
-	for _, key := range *serverCache {
-		serverPSK = key
-	}
-	var clientPSK PreSharedKey
-	for _, key := range *clientCache {
-		clientPSK = key
-	}
-	assertDeepEquals(t, clientPSK, serverPSK)
```
bifurcation: Even though the PSKs won't match exactly, we should check that the parts that should be the same are the same.  Could you please at least check that the identity and key are the same?

handshake.go
```diff
@@ -157,6 +158,9 @@ func (h *ClientHandshake) HandleNewSessionTicket(hm *HandshakeMessage) (PreShare
 		IsResumption: true,
 		Identity:     tkt.Ticket,
 		Key:          h.Context.resumptionSecret,
+		ReceivedAt:   time.Now(),
+		ExpiresAt:    time.Now().Add(time.Duration(tkt.TicketLifetime) * time.Second),
```
bifurcation: You should check ExpiresAt when you pull the ticket and if it's expired, don't send it.  It would also be good to go ahead and delete it, but you'll need to add a `Delete(string)` method to `ClientSessionCache` for that.

bifurcation:
Replaced by #126

PR #119, opened 2017-01-24
Store ALPN in the session cache

ekr:
@bifurcation this is causing failures.

You have a bigger problem, though, which is that if ALPN doesn't match you need to reject resumption, not fail the handshake, but this will at least allow resumption.


bifurcation:
CircleCI is busted; merging based on manual review.

bifurcation:
Actually not merging, because the Github UI won't let me (?)

bifurcation:
@ekr It looks like I've gotten CircleCI working again.  Could you bump this or resubmit to kick it off?

bifurcation:
This seems to be fixed now.

https://github.com/bifurcation/mint/blob/master/negotiation.go#L178

PR #127, opened 2017-06-06
Big refactor of the I/O discipline. The big change is that reading

ekr:
is a lot less greedy.

- Build a generic framed data state machine used by both the
  record and the handshake layers.

- Allow Read() to return whenever any data is available rather
  than when all the data is available.

- The frameReader can return frameWouldBlock when it tries
  to read off a reader and gets 0 bytes. This is handled
  in the upper layers by looping and controlled by a Nonblocking
  config parameter.

This code doesn't support nonblocking at the top-level API yet,
but I'll be adding that next.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -74,6 +74,7 @@ type Config struct {
 	NextProtos       []string
 	PSKs             PreSharedKeyCache
 	PSKModes         []PSKKeyExchangeMode
+	NonBlocking      bool
 
 	// The same config object can be shared among different connections, so it
 	// needs its own mutex
@@ -185,6 +186,7 @@ type Conn struct {
 	EarlyData []byte
 
 	state             StateConnected
+	hState            HandshakeState
 	handshakeMutex    sync.Mutex
 	handshakeAlert    Alert
 	handshakeComplete bool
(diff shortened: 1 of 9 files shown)


frame-reader.go
```diff
@@ -0,0 +1,150 @@
...
+
+var frameReaderWouldBlock = fmt.Errorf("Would have blocked")
+
+type frameNextAction func(f *frameReader) error
+
+type frameReader struct {
+	conn        io.Reader
+	details     frameDetails
+	state       uint8
+	hdr         []byte
```
bifurcation: Nit: Would prefer this be spelled out, maybe even `cachedHeader`

frame-reader.go
```diff
@@ -0,0 +1,150 @@
...
+func (f *frameReader) readChunk() (hdr []byte, body []byte, err error) {
+	var buf []byte
+
+	// Loop until one of three things happens:
+	//
+	// 1. We process a record
+	// 2. We try to read off the socket and get nothing, in which case
+	//    return frameReaderWouldBlock
+	// 3. We get an error.
+	//
```
bifurcation: Nit: Extra line

frame-reader.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (f *frameReader) needed() int {
+	tmp := (len(f.working) - f.writeOffset) - len(f.remainder)
+	if tmp < 0 {
+		return 0
+	}
+	return tmp
+}
+
+func (f *frameReader) readChunk() (hdr []byte, body []byte, err error) {
```
bifurcation: Nit: `readFrame` (since you always return a whole frame)

frame-reader.go
```diff
@@ -0,0 +1,150 @@
+// Read a generic "framed" packet consisting of a header and a
+// This is used for both TLS Records and TLS Handshake Messages
+package mint
+
+import (
+	"fmt"
+	"io"
+)
+
+type frameDetails interface {
```
bifurcation: Nit: I might prefer if this were named something related to "framing", since you're describing a framing scheme, instead of a frame.  Maybe just call it `framing`?

frame-reader.go
```diff
@@ -0,0 +1,150 @@
...
+	kFrameReaderHdr  = 0
+	kFrameReaderBody = 1
+)
+
+var frameReaderWouldBlock = fmt.Errorf("Would have blocked")
+
+type frameNextAction func(f *frameReader) error
+
+type frameReader struct {
+	conn        io.Reader
```
bifurcation: It seems like right now there are two ways for data to come into the `frameReader`, (1) from `conn`and (2) from `addChunk`.  It seems like you could potentially simplify this by getting rid of the `conn` option.  You would just be requiring the caller to explicitly pass chunks in from its reader.  But then you could probably simplify some of the logic here and fold it into addChunk (i.e., just have `addChunk` find all the complete frames in the extended byffer and either return them or buffer them).   Removing the `conn` option would also remove all the concerns around blocking.

```
for !fr.HasFrames() {
    in.Read(buf) // <-- wrap blocking logic around this
    fr.AddChunk(buf)
}
```

conn.go
```diff
@@ -185,6 +186,7 @@ type Conn struct {
 	EarlyData []byte
 
 	state             StateConnected
+	hState            HandshakeState
```
ekr: I'm not sure I follow this suggestion.

frame-reader.go
ekr: I did some of this.

I don't think you want to try to find *all* the frames because returning a bunch of frames is hard to manage, and you still need to know whether more data is needed to complete some partial frame. So, I did the first but not the second.

frame-reader.go
ekr: Not needed.

PR #129, opened 2017-06-15
Make the SendSessionTicket configuration parameter actually work.

ekr:
Note that this is an API change in that the default in the
configuration is for this value to be false. Thus, people who
were previously relying on the default (broken) behavior for
this configuration flag will not get session tickets. Alternately,
we can remove this config flag and replace it with one that
goes the other way.


ekr:
Something is actually wrong with this. Do not merge.

ekr:
Grr.... Problem in MINQ.

PR #130, opened 2017-06-19
Implement 1-RTT Exporter.

ekr:


ekr:
@bifurcation: I think this is ready but I'd like a review from @martinthomson first.

server-state-machine.go
```diff
@@ -785,6 +800,7 @@ type ServerStateWaitFinished struct {
 	handshakeHash       hash.Hash
 	clientTrafficSecret []byte
 	serverTrafficSecret []byte
+	exporterSecret      []byte
```
martinthomson: The duplication here is starting to look pretty bad.  Any reason why these state variables can't be bound up into a struct that can be reused?  This is probably a question for @bifurcation who created this miss in the first place.

conn.go
```diff
@@ -729,3 +727,20 @@ func (c *Conn) SendKeyUpdate(requestUpdate bool) error {
 func (c *Conn) GetHsState() string {
 	return reflect.TypeOf(c.hState).Name()
 }
+
+func (c *Conn) ComputeExporter(label string, context []byte, keyLength int) ([]byte, error) {
+	_, ok := c.hState.(StateConnected)
+	if !ok {
```
martinthomson: s/ok/connected/, as above

conn.go
```diff
@@ -729,3 +727,20 @@ func (c *Conn) SendKeyUpdate(requestUpdate bool) error {
 func (c *Conn) GetHsState() string {
 	return reflect.TypeOf(c.hState).Name()
 }
+
+func (c *Conn) ComputeExporter(label string, context []byte, keyLength int) ([]byte, error) {
```
martinthomson: Just to confirm: the plan is to add a `ComputeEarlyExporter` function for the early exporter and then to factor out the necessary parts of this function.

conn.go
martinthomson: That said, if you want to factor out the early exporter, then checking that the appropriate secret is non-nil will achieve the same goal.

conn.go
martinthomson: I just learned something about go today.  Apparently `string` is [essentially the same](https://blog.golang.org/strings) as `[]byte`.  I don't know what to do about that though.

conn_test.go
```diff
@@ -274,6 +274,12 @@ func assertKeySetEquals(t *testing.T, k1, k2 keySet) {
 	assertByteEquals(t, k1.key, k2.key)
 }
 
+func computeExporter(t *testing.T, c *Conn, label string, length int) []byte {
+	res, err := c.ComputeExporter(label, []byte{}, length)
```
martinthomson: The tests should exercise non-zero-length contexts.

conn.go
ekr: yeah

conn_test.go
ekr: I didn't expose non-zero length contexts. Do you think we need them?

ekr: Oh, never mind. Am idiot.

bifurcation:
Replaced by #135

PR #133, opened 2017-07-08
Support draft-21

bifurcation:
This just changes the version number and adds support for ticket nonces, which should be enough.  

However, draft-20 support never really got interop tested, so let's do some testing with other stacks before merging.

### DIFF
--- a/common.go
+++ b/common.go
@@ -1,7 +1,7 @@
 package mint
 
 var (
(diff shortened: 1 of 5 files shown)


bifurcation:
Merging based on bidirectional interop with the -21 implementation in OpenSSL master.

PR #135, opened 2017-07-14
Minq draft 21

ekr:


handshake-layer.go
```diff
@@ -173,10 +173,12 @@ func (h *HandshakeLayer) sendAlert(err Alert) error {
 
 func (h *HandshakeLayer) ReadMessage() (*HandshakeMessage, error) {
 	var hdr, body []byte
-	err := WouldBlock
+	err := error(nil)
```
marten-seemann: Better: `var err error`

conn.go
```diff
@@ -728,3 +729,33 @@ func (c *Conn) SendKeyUpdate(requestUpdate bool) error {
...
+	}
+
+	h0 := c.state.cryptoParams.Hash.New().Sum(nil)
+	tmpSecret := deriveSecret(c.state.cryptoParams, c.state.exporterSecret, label, h0)
+
+	hc := c.state.cryptoParams.Hash.New().Sum(context)
+	return HkdfExpandLabel(c.state.cryptoParams.Hash, tmpSecret, "exporter", hc, keyLength), nil
+}
+
+func (c *Conn) GetConnectionState() ConnectionState {
```
bifurcation: Nit: Think `State()` would be more aesthetic

conn.go
```diff
@@ -173,9 +173,10 @@ var (
 )
 
 type ConnectionState struct {
-	HandshakeComplete bool                // TLS handshake is complete
-	CipherSuite       CipherSuite         // cipher suite in use (TLS_RSA_WITH_RC4_128_SHA, ...)
-	PeerCertificates  []*x509.Certificate // certificate chain presented by remote peer
+	HandshakeState   string              // string representation of the handshake state.
+	CipherSuite      CipherSuiteParams   // cipher suite in use (TLS_RSA_WITH_RC4_128_SHA, ...)
+	PeerCertificates []*x509.Certificate // certificate chain presented by remote peer TODO(ekr@rtfm.com): implement
+	NextProto        string              // Selected ALPN proto
```
bifurcation: This is starting to look mighty similar to [`ConnectionParams`](https://github.com/bifurcation/mint/blob/master/state-machine.go#L70).  Is there some way to de-dup?

frame-reader.go
```diff
@@ -2,8 +2,7 @@
 // This is used for both TLS Records and TLS Handshake Messages
 package mint
 
-import (
-)
+import ()
```
bifurcation: Nit: You can just delete this line.

conn.go
ekr: Probably but I think too hard for this PR. See #136

PR #137, opened 2017-09-06
WIP: external extensions

ekr:
So you can see, not for merging

### DIFF
--- a/state-machine.go
+++ b/state-machine.go
@@ -36,6 +36,11 @@ type HandshakeState interface {
 	Next(hm *HandshakeMessage) (HandshakeState, []HandshakeAction, Alert)
 }
 
+type AppExtensionHandler interface {
+	Send(hs HandshakeType, el *ExtensionList) error
+	Receive(hs HandshakeType, el *ExtensionList) error
+}
+
 // Capabilities objects represent the capabilities of a TLS client or server,
 // as an input to TLS negotiation
 type Capabilities struct {
@@ -46,6 +51,7 @@ type Capabilities struct {
 	PSKs             PreSharedKeyCache
 	Certificates     []*Certificate
 	AuthCertificate  func(chain []CertificateEntry) error
+	ExtensionHandler AppExtensionHandler
 
 	// For client
 	PSKModes []PSKKeyExchangeMode
--- a/client-state-machine.go
+++ b/client-state-machine.go
@@ -132,6 +132,15 @@ func (state ClientStateStart) Next(hm *HandshakeMessage) (HandshakeState, []Hand
 		}
 	}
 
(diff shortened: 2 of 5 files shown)


ekr:
@bifurcation: PTAL

state-machine.go
```diff
@@ -36,6 +36,11 @@ type HandshakeState interface {
 	Next(hm *HandshakeMessage) (HandshakeState, []HandshakeAction, Alert)
 }
 
+type AppExtensionHandler interface {
+	Send(hs HandshakeType, el *ExtensionList) error
+	Receive(hs HandshakeType, el *ExtensionList) error
+}
```
marten-seemann: I think a more readable interface would be
```go
Send(HandshakeType) ([]Extension, error)
Receive(HandshakeType, Extension) (error)
```
i.e. `Send` (maybe `Get` or `GetExtensions` would be better) returns a slice of extensions to add, and `Receive` (or `Received`) is called separately for every extension received.

ekr:
Yeah, this is kind of a defect in Go, In C++, of course, we would just have Receive() take a const ExtensionList&. 

I'm a bit reluctant to use the interface you suggest, for several reasons:

1. It's more disruptive to the general style, which is that one .Add()s to the extension list.
2. It's less symmetrical.

It's also slightly less efficient, as you'll need to make a copy of the list on return.

marten-seemann:
1. I've been working on a PR refactoring the `ExtensionList`. I think the API would be cleaner if we didn't have `Add` and `Find` (by using a map to represent the extension list instead of a slice), but I'm still working on the details of that.
2. I'd argue that the asymmetry is due to what we're actually doing here: In the first case, we're generating new extensions, and in the second case we're consuming the extension data). With my proposed interface would allow to only pass the external extensions to the registered extension handler.

I'd propose this: I'll hold back my PR refactoring the extensions until this is merged, and we can discuss my changes for the extension handling as a whole. What do you think?

ekr:
That works for me, but let's hear from @rbarnes.

bifurcation:
@marten-seemann The trade-off here is that representing it as a slice makes marshal/unmarshal [trivial](https://github.com/bifurcation/mint/blob/master/extensions.go#L25).  So you have to choose whether to have the complexity in the marshal/unmarshal code or in the add/find code.  Personally, I would prefer to keep the marshal/unmarshal simple, but *de gustibus...*.

Nonetheless, the plan to land this first and then revisit the question of how to represent extensions seems sensible to me.

ekr:
@barnes: let's merge this as-is for now. I think it's otherwise ready

bifurcation:
SGTM.  Merging now.

PR #138, opened 2017-09-06
make go vet happy and run gofmt -s

marten-seemann:


conn.go
```diff
@@ -83,6 +83,35 @@ type Config struct {
 	mutex sync.RWMutex
 }
 
+// Clone returns a shallow clone of c. It is safe to clone a Config that is
+// being used concurrently by a TLS client or server.
+func (c *Config) Clone() *Config {
```
bifurcation: In general, prefer that unrelated changes be in separate PRs.

PR #139, opened 2017-09-07
implement a CookieHandler to manually set and validate cookies

marten-seemann:
By setting a `CookieHandler`, the user can provide an opaque byte-slice that is used as a cookie in the HRR, and validate the cookie when the client echos it in the CH. Mint passes a pointer to the `Conn` to the methods that generate and validate the cookie. It's the responsibility of the cookie handler implementation to decide if a cookie should be sent for this specific connection..

If `Config.RequireCookie` is set to `true`, but no `Config.CookieHandler` is defined, a default cookie handler is used (the one that was already implemented). The default handler uses 32 random bytes as a cookie.

I'm not sure where the appropriate place would be to test a failed cookie validation. I guess *state-machine_test.go* would be the right place, but the tests there only cover successful handshakes.

@ekr: It would be great if you could review this PR as well.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -39,11 +39,11 @@ type PreSharedKeyCache interface {
 type PSKMapCache map[string]PreSharedKey
 
 // A CookieHandler does two things:
-// - generates cookies that are sent to the client in the HelloRetryRequest
-// - validates cookies sent by the client in a ClientHello
+// - generates a byte string that is sent as a part of a cookie to the client in the HelloRetryRequest
+// - validates this byte string echoed by the client in the ClientHello
 type CookieHandler interface {
-	Generate(*Conn) (*CookieExtension, error)
-	Validate(*Conn, *CookieExtension) bool
+	Generate(*Conn) ([]byte, error)
+	Validate(*Conn, []byte) bool
 }
 
 func (cache PSKMapCache) Get(key string) (psk PreSharedKey, ok bool) {
--- a/conn.go
+++ b/conn.go
@@ -38,6 +38,14 @@ type PreSharedKeyCache interface {
 
 type PSKMapCache map[string]PreSharedKey
 
+// A CookieHandler does two things:
+// - generates cookies that are sent to the client in the HelloRetryRequest
+// - validates cookies sent by the client in a ClientHello
+type CookieHandler interface {
+	Generate(*Conn) (*CookieExtension, error)
+	Validate(*Conn, *CookieExtension) bool
+}
+
 func (cache PSKMapCache) Get(key string) (psk PreSharedKey, ok bool) {
 	psk, ok = cache[key]
 	return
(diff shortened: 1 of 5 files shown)


ekr:
Hey, that's great. This was coming up soon on my list. I'm on travel today, but I'll take a look as soon as I can.

ekr:
@marten-seemann, @rbarnes: I think there's an underlying problem in Mint, which is that it generates HRR when RequireCookie is set but not when it is missing the appropriate client key share. So, that's going to cause problems later. You might look at MT's code in NSS for handling this that lets you cover the cases.

marten-seemann:
This PR is ready for review now.

I updated the description to explain the changes.

@ekr: I think the HRR generation is a general mint problem, and unrelated to this PR. Can you create a new issue for this?

marten-seemann:
@bifurcation: Can you please review and merge this PR?

conn.go
```diff
@@ -38,6 +38,14 @@ type PreSharedKeyCache interface {
 
 type PSKMapCache map[string]PreSharedKey
 
+// A CookieHandler does two things:
+// - generates cookies that are sent to the client in the HelloRetryRequest
+// - validates cookies sent by the client in a ClientHello
+type CookieHandler interface {
+	Generate(*Conn) (*CookieExtension, error)
+	Validate(*Conn, *CookieExtension) bool
```
ekr: Can you add a comment on what Validate's returns are?

conn.go
```diff
@@ -64,8 +72,12 @@ type Config struct {
 	TicketLen          int
 	EarlyDataLifetime  uint32
 	AllowEarlyData     bool
-	RequireCookie      bool
-	RequireClientAuth  bool
+	// Require the client to echo a cookie.
+	RequireCookie bool
+	// If cookies are required and no CookieHandler is set, a default cookie handler is used.
+	// The default cookie handler uses a 32 random bytes as a cookie.
```
ekr: Nit: either "32 random bytes" or "a random 32-byte value"

server-state-machine.go
```diff
@@ -218,7 +219,8 @@ func (state ServerStateStart) Next(hm *HandshakeMessage) (HandshakeState, []Hand
 
 		nextState := ServerStateStart{
 			Caps:              state.Caps,
-			cookie:            cookie.Cookie,
+			conn:              state.conn,
+			cookieSent:        true,
```
ekr: Why did you move cookie into conn rather than leaving it in ServerStateStrart?

state-machine_test.go
```diff
@@ -70,6 +70,7 @@ var (
 				PSKs:             &PSKMapCache{},
 				Certificates:     certificates,
 				RequireCookie:    true,
+				CookieHandler:    &defaultCookieHandler{},
```
ekr: Is this needed? It seems like you do this by default. Wouldn't a better test be to show you could override the cookie handler.

marten-seemann: The `conn` sets the default value, but not the `ServerStateStart`. There are a bunch of tests I'd like here (at least: default cookie handler, custom cookie handler, and failed cookie validation), but the current format of the tests doesn't really allow to perform those.

server-state-machine.go
marten-seemann: The `conn` is needed as a parameter for `CookieHandler.Generate` and `Validate` anyway.

conn.go
marten-seemann: I thought this is covered by the comment for the interface. Which additional information would you like here?

conn.go
marten-seemann: Fixed.

marten-seemann:
@ekr: Please have another look.

marten-seemann:
I added another fix here. If the `CookieHandler` doesn't return a cookie, mint shouldn't send one.

This should be ready for merging now. @ekr, @bifurcation, what do you think?

PR #140, opened 2017-09-08
implement a string representation of the CipherSuite

marten-seemann:
I'd like to be able to log the cipher suite negotiated in the handshake. Having a string representation of the `CipherSuite` would be very handy for that.

### DIFF
--- a/common.go
+++ b/common.go
@@ -1,5 +1,10 @@
 package mint
 
+import (
+	"fmt"
+	"strconv"
+)
+
 var (
 	supportedVersion uint16 = 0x7f15 // draft-21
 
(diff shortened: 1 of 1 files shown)


common.go
```diff
@@ -51,6 +53,24 @@ const (
...
+	case TLS_AES_256_GCM_SHA384:
+		return "TLS_AES_256_GCM_SHA384"
+	case TLS_CHACHA20_POLY1305_SHA256:
+		return "TLS_CHACHA20_POLY1305_SHA256"
+	case TLS_AES_128_CCM_SHA256:
+		return "TLS_AES_128_CCM_SHA256"
+	case TLS_AES_256_CCM_8_SHA256:
+		return "TLS_AES_256_CCM_8_SHA256"
+	}
+	return fmt.Sprintf("invalid CipherSuite value")
```
ekr: Did you mean to be passing back the value in this answer? If not, why are you bothering with Sprintf?

marten-seemann: That was the plan, but `%x` leads to an infinite recursion here. I'll find a workaround for that.

marten-seemann: Done.

PR #141, opened 2017-11-08
implement stateless resets

marten-seemann:
This PR changes the server state machine such that sending HelloRetryRequests can be performed statelessly.

In non-blocking mode, `Handshake()` will return a new alert, `AlertStatelessRetry`. When receiving this return, an app can discard all mint-related state (i.e. delete `Conn` garbage collected). When not using non-blocking mode, the handshake is performed as stateless as possible, at no point the initial ClientHello and the HelloRetryRequest will be stored.

The cookie that is sent to the client in the HelloRetryRequest is composed of two parts: In the mint-part of the cookie, we store the hash of the initial ClientHello. The second part can be used to store an application-defined cookie (used when `Config.CookieHandler` is set). The application has to provide a way to encrypt this cookie (it should use authenticated encryption for this). This can be done by setting `Config.CookieSource`. In blocking mode, we default to use AES-GCM if no `CookieSource` is set.

@ekr, @bifurcation: Please tell me what you think about this. The stateless handshake works for me with this PR, but I can't guarantee I didn't miss anything subtle here. My current understanding of the TLS 1.3 draft doesn't go a lot farther than what I need for my mint changes and the QUIC integration.

### DIFF
--- a/alert.go
+++ b/alert.go
@@ -46,6 +46,7 @@ const (
 	AlertBadCertificateHashValue     Alert = 114
 	AlertUnknownPSKIdentity          Alert = 115
 	AlertNoApplicationProtocol       Alert = 120
+	AlertStatelessRetry              Alert = 253
 	AlertWouldBlock                  Alert = 254
 	AlertNoAlert                     Alert = 255
 )
@@ -82,6 +83,7 @@ var alertText = map[Alert]string{
 	AlertUnknownPSKIdentity:          "unknown PSK identity",
 	AlertNoApplicationProtocol:       "no application protocol",
 	AlertNoRenegotiation:             "no renegotiation",
+	AlertStatelessRetry:              "stateless retry",
 	AlertWouldBlock:                  "would have blocked",
 	AlertNoAlert:                     "no alert",
 }
--- a/conn.go
+++ b/conn.go
@@ -36,15 +36,19 @@ type PreSharedKeyCache interface {
 	Size() int
 }
 
-type PSKMapCache map[string]PreSharedKey
-
-// A CookieHandler does two things:
-// - generates a byte string that is sent as a part of a cookie to the client in the HelloRetryRequest
-// - validates this byte string echoed by the client in the ClientHello
+// A CookieHandler can be used to give the application more fine-grained control over Cookies.
+// Generate receives the Conn as an argument, so the CookieHandler can decide when to send the cookie based on that, and offload state to the client by encoding that into the Cookie.
+// When the client echoes the Cookie, Validate is called. The application can then recover the state from the cookie.
 type CookieHandler interface {
+	// Generate a byte string that is sent as a part of a cookie to the client in the HelloRetryRequest
+	// If Generate returns nil, mint will not send a HelloRetryRequest.
 	Generate(*Conn) ([]byte, error)
+	// Validate is called when receiving a ClientHello containing a Cookie.
+	// If validation failed, the handshake is aborted.
 	Validate(*Conn, []byte) bool
 }
+
+type PSKMapCache map[string]PreSharedKey
 
 func (cache PSKMapCache) Get(key string) (psk PreSharedKey, ok bool) {
 	psk, ok = cache[key]
@@ -74,9 +78,16 @@ type Config struct {
 	AllowEarlyData     bool
 	// Require the client to echo a cookie.
 	RequireCookie bool
-	// If cookies are required and no CookieHandler is set, a default cookie handler is used.
-	// The default cookie handler uses 32 random bytes as a cookie.
-	CookieHandler     CookieHandler
+	// A CookieHandler can be used to set and validate a cookie.
+	// The cookie returned by the CookieHandler will be part of the cookie sent on the wire, and encoded using the CookieSource.
+	// If no CookieHandler is set, mint will always send a cookie.
+	// The CookieHandler can be used to decide on a per-connection basis, if a cookie should be sent.
+	CookieHandler CookieHandler
+	// The CookieSource is used to encrypt / decrypt cookies.
+	// It should make sure that the Cookie cannot be read and tampered with by the client.
+	// If non-blocking mode is used, and cookies are required, this field has to be set.
+	// In blocking mode, a default cookie source is used, if this is unused.
+	CookieSource      CookieSource
 	RequireClientAuth bool
 
 	// Shared fields
@@ -110,6 +121,8 @@ func (c *Config) Clone() *Config {
 		EarlyDataLifetime:  c.EarlyDataLifetime,
 		AllowEarlyData:     c.AllowEarlyData,
 		RequireCookie:      c.RequireCookie,
+		CookieHandler:      c.CookieHandler,
+		CookieSource:       c.CookieSource,
 		RequireClientAuth:  c.RequireClientAuth,
 
 		Certificates:     c.Certificates,
@@ -611,6 +624,7 @@ func (c *Conn) HandshakeSetup() Alert {
 		PSKModes:          c.config.PSKModes,
 		AllowEarlyData:    c.config.AllowEarlyData,
 		RequireCookie:     c.config.RequireCookie,
+		CookieSource:      c.config.CookieSource,
 		CookieHandler:     c.config.CookieHandler,
 		RequireClientAuth: c.config.RequireClientAuth,
 		NextProtos:        c.config.NextProtos,
@@ -621,10 +635,6 @@ func (c *Conn) HandshakeSetup() Alert {
 		ServerName: c.config.ServerName,
 		NextProtos: c.config.NextProtos,
 		EarlyData:  c.EarlyData,
-	}
-
-	if caps.RequireCookie && caps.CookieHandler == nil {
-		caps.CookieHandler = &defaultCookieHandler{}
 	}
 
 	if c.isClient {
@@ -642,6 +652,19 @@ func (c *Conn) HandshakeSetup() Alert {
 			}
 		}
 	} else {
+		if c.config.RequireCookie && c.config.CookieSource == nil {
+			logf(logTypeHandshake, "RequireCookie set, but no CookieSource provided. Using default cookie source. Stateless Retry not possible.")
+			if c.config.NonBlocking {
+				logf(logTypeHandshake, "Not possible in non-blocking mode.")
+				return AlertInternalError
+			}
+			var err error
+			caps.CookieSource, err = newDefaultCookieSource()
+			if err != nil {
+				logf(logTypeHandshake, "Error initializing cookie source: %v", alert)
+				return AlertInternalError
+			}
+		}
 		state = ServerStateStart{Caps: caps, conn: c}
 	}
 
@@ -671,7 +694,7 @@ func (c *Conn) Handshake() Alert {
 
 	var alert Alert
 	if c.hState == nil {
-		logf(logTypeHandshake, "%s First time through handshake, setting up", label)
+		logf(logTypeHandshake, "%s First time through handshake (or after stateless retry), setting up", label)
 		alert = c.HandshakeSetup()
 		if alert != AlertNoAlert {
 			return alert
@@ -701,16 +724,14 @@ func (c *Conn) Handshake() Alert {
 
 		// Advance the state machine
 		state, actions, alert = state.Next(hm)
-
-		if alert != AlertNoAlert {
+		if alert != AlertNoAlert && alert != AlertStatelessRetry {
 			logf(logTypeHandshake, "Error in state transition: %v", alert)
 			return alert
 		}
 
 		for index, action := range actions {
 			logf(logTypeHandshake, "%s taking next action (%d)", label, index)
-			alert = c.takeAction(action)
-			if alert != AlertNoAlert {
+			if alert := c.takeAction(action); alert != AlertNoAlert {
 				logf(logTypeHandshake, "Error during handshake actions: %v", alert)
 				c.sendAlert(alert)
 				return alert
@@ -719,7 +740,6 @@ func (c *Conn) Handshake() Alert {
 
 		c.hState = state
 		logf(logTypeHandshake, "state is now %s", c.GetHsState())
-
 		_, connected = state.(StateConnected)
 	}
 
--- a/conn.go
+++ b/conn.go
@@ -79,15 +79,15 @@ type Config struct {
 	// Require the client to echo a cookie.
 	RequireCookie bool
 	// A CookieHandler can be used to set and validate a cookie.
-	// The cookie returned by the CookieHandler will be part of the cookie sent on the wire, and encoded using the CookieSource.
+	// The cookie returned by the CookieHandler will be part of the cookie sent on the wire, and encoded using the CookieProtector.
 	// If no CookieHandler is set, mint will always send a cookie.
 	// The CookieHandler can be used to decide on a per-connection basis, if a cookie should be sent.
 	CookieHandler CookieHandler
-	// The CookieSource is used to encrypt / decrypt cookies.
+	// The CookieProtector is used to encrypt / decrypt cookies.
 	// It should make sure that the Cookie cannot be read and tampered with by the client.
 	// If non-blocking mode is used, and cookies are required, this field has to be set.
-	// In blocking mode, a default cookie source is used, if this is unused.
-	CookieSource      CookieSource
+	// In blocking mode, a default cookie protector is used, if this is unused.
+	CookieProtector   CookieProtector
 	RequireClientAuth bool
 
 	// Shared fields
@@ -122,7 +122,7 @@ func (c *Config) Clone() *Config {
 		AllowEarlyData:     c.AllowEarlyData,
 		RequireCookie:      c.RequireCookie,
 		CookieHandler:      c.CookieHandler,
-		CookieSource:       c.CookieSource,
+		CookieProtector:    c.CookieProtector,
 		RequireClientAuth:  c.RequireClientAuth,
 
 		Certificates:     c.Certificates,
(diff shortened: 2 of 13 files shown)


ekr:
@marten-seemann this seems to fail in CI...

alert.go
```diff
@@ -82,6 +83,7 @@ var alertText = map[Alert]string{
 	AlertUnknownPSKIdentity:          "unknown PSK identity",
 	AlertNoApplicationProtocol:       "no application protocol",
 	AlertNoRenegotiation:             "no renegotiation",
+	AlertStatelessRetry:              "stateless retry",
```
ekr: This isn't your fault, but I wonder if we should give these non-alerts a special name so we don't acccidentally try to send them.

conn.go
```diff
@@ -74,9 +74,16 @@ type Config struct {
 	AllowEarlyData     bool
 	// Require the client to echo a cookie.
 	RequireCookie bool
+	// TODO: fix comment here: no more default cookie handler
```
ekr: This TODO seems like it needs fixing.

conn.go
```diff
@@ -74,9 +74,16 @@ type Config struct {
...
 	// If cookies are required and no CookieHandler is set, a default cookie handler is used.
 	// The default cookie handler uses 32 random bytes as a cookie.
-	CookieHandler     CookieHandler
+	CookieHandler CookieHandler
+	// The CookieSource is used to encrypt / decrypt cookies.
+	// TODO: implement a check for that
+	// If non-blocking mode is used, and cookies are required, this field has to be set.
+	// TODO: implement that
```
ekr: These TODOs also seem to need fixing

conn.go
```diff
@@ -642,6 +648,19 @@ func (c *Conn) HandshakeSetup() Alert {
...
+		if c.config.RequireCookie && c.config.CookieSource == nil {
+			logf(logTypeHandshake, "RequireCookie set, but no CookieSource provided. Using default cookie source. Stateless Retry not possible.")
+			if c.config.NonBlocking {
+				logf(logTypeHandshake, "Not possible in non-blocking mode.")
+				return AlertInternalError
+			}
+			var err error
+			caps.CookieSource, err = newDefaultCookieSource()
+			if err != nil {
+				logf(logTypeHandshake, "Error initializing client state: %v", alert)
```
ekr: Should this say server state?

conn.go
```diff
@@ -687,6 +706,7 @@ func (c *Conn) Handshake() Alert {
 
 	for !connected {
 		// Read a handshake message
+		logf(logTypeHandshake, "readmessage")
```
ekr: Maybe make this |logTypeVerbose| and make the text clearer.

conn.go
```diff
@@ -720,6 +738,13 @@ func (c *Conn) Handshake() Alert {
 		c.hState = state
 		logf(logTypeHandshake, "state is now %s", c.GetHsState())
 
+		// if alert == AlertStatelessRetry {
+		// 	logf(logTypeHandshake, "Doing a retry")
+		// 	if c.config.NonBlocking {
+		// 		return AlertStatelessRetry
+		// 	}
+		// }
```
ekr: You should remove this code.

cookie_source.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func deriveKey(secret []byte) ([]byte, error) {
+	r := hkdf.New(sha256.New, secret, nil, []byte("mint TLS 1.3 cookie token key"))
+	key := make([]byte, tokenKeySize)
+	if _, err := io.ReadFull(r, key); err != nil {
+		return nil, err
+	}
+	return key, nil
+}
```
ekr: I'm not sure it's necessary to have an HKDF stage here. You are feeding in random bits, right?

cookie_source.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	return &defaultCookieSource{aead: aead}, nil
+}
+
+func (s *defaultCookieSource) NewToken(data []byte) ([]byte, error) {
+	nonce := make([]byte, tokenNonceSize)
+	if _, err := rand.Read(nonce); err != nil {
+		return nil, err
+	}
+	return s.aead.Seal(nonce, nonce, data, nil), nil
```
ekr: I don't think this is the crypto structure you want here. The issue is that you have a non-trivial chance of nonce-reuse at 2^96 because of birthday, which is catastrophic with GCM. You should either use an integer nonce (that won't repeat within any reasonable bound) or instead generate a longer nonce (256 bits is good) and use it to derive both the per-message key and per-message nonce

server-state-machine.go
```diff
@@ -57,13 +60,18 @@ import (
...
 //  CONNECTED					StoreTicket || (RekeyIn; [RekeyOut])
 
+// A cookie can be sent to the client in a HRR.
+// It contains two fields:
+// 1. The MintCookie: This field is used by mint itself to store the hash of initial client hello.
+// 2. The ApplicationCookie: This opaque value can be provided by the application (by setting a Config.CookieHandler)
+type cookie struct {
+	MintCookie        []byte `tls:"head=2"`
+	ApplicationCookie []byte `tls:"head=2"`
+}
```
ekr: In NSS, we also start the negotiated parameters and double-check that they don't change

server-state-machine.go
```diff
@@ -129,9 +139,29 @@ func (state ServerStateStart) Next(hm *HandshakeMessage) (HandshakeState, []Hand
...
-	if state.Caps.RequireCookie && state.cookieSent && !state.Caps.CookieHandler.Validate(state.conn, clientCookie.Cookie) {
-		logf(logTypeHandshake, "[ServerStateStart] Cookie mismatch")
-		return nil, nil, AlertAccessDenied
+	// The client sent a cookie. So this is probably the second ClientHello (sent as a response to a HRR)
+	var firstClientHello *HandshakeMessage
+	if clientSentCookie {
+		plainCookie, err := state.Caps.CookieSource.DecodeToken(clientCookie.Cookie)
+		if err != nil {
+			logf(logTypeHandshake, fmt.Sprintf("[ServerStateStart] Error decoding token [%v]", err))
+			return nil, nil, AlertAccessDenied
```
ekr: I don't think you want AccessDenied here, but rather DecryptError or IllegalParameter

server-state-machine.go
```diff
@@ -129,9 +139,29 @@ func (state ServerStateStart) Next(hm *HandshakeMessage) (HandshakeState, []Hand
...
+		}
+		cookie := &cookie{}
+		if _, err := syntax.Unmarshal(plainCookie, cookie); err != nil {
+			logf(logTypeHandshake, fmt.Sprintf("[ServerStateStart] Error unmarshaling cookie [%v]", err))
+			return nil, nil, AlertAccessDenied
+		}
```
ekr: This should be internal_error, because it can't happen.

server-state-machine.go
ekr: I would put a TODO here because if you were to change your key, this wouldn't work.

server-state-machine.go
```diff
@@ -142,26 +172,51 @@ func (state ServerStateStart) Next(hm *HandshakeMessage) (HandshakeState, []Hand
...
-		}
 
-		chTrunc, err := ch.Truncated()
+	if len(clientPSK.Identities) > 0 {
+		canDoPSK, selectedPSK, psk, params, err = PSKNegotiation(clientPSK.Identities, state.Caps.PSKs)
 		if err != nil {
-			logf(logTypeHandshake, "[ServerStateStart] Error computing truncated ClientHello [%v]", err)
-			return nil, nil, AlertDecodeError
+			logf(logTypeHandshake, "[ServerStateStart] Error in PSK negotiation [%v]", err)
+			return nil, nil, AlertInternalError
```
ekr: I am surprised this is internal error. @bifurcation ?

server-state-machine.go
```diff
@@ -142,26 +172,51 @@ func (state ServerStateStart) Next(hm *HandshakeMessage) (HandshakeState, []Hand
...
+			if err != nil {
+				logf(logTypeHandshake, "[ServerStateStart] Error computing truncated ClientHello [%v]", err)
+				return nil, nil, AlertDecodeError
+			}
+			ctxHash.Write(chTrunc)
+
+			binder := computeFinishedData(params, binderKey, ctxHash.Sum(nil))
+			if !bytes.Equal(binder, clientPSK.Binders[selectedPSK].Binder) {
+				logf(logTypeNegotiation, "Binder check failed for identity %x; [%x] != [%x]", psk.Identity, binder, clientPSK.Binders[selectedPSK].Binder)
+				return nil, nil, AlertInternalError
```
ekr: This seems like decrypt_error.

server-state-machine.go
```diff
@@ -175,61 +230,66 @@ func (state ServerStateStart) Next(hm *HandshakeMessage) (HandshakeState, []Hand
...
+			if state.Caps.CookieHandler == nil { // if Config.RequireCookie is set, but no CookieHandler was provided, we definitely need to send a cookie
+				shouldSendHRR = true
+			} else { // if the CookieHandler was set, we just send a cookie when the application provides one
+				var err error
+				appCookie, err = state.Caps.CookieHandler.Generate(state.conn)
+				if err != nil {
+					logf(logTypeHandshake, "[ServerStateStart] Error generating cookie [%v]", err)
+					return nil, nil, AlertInternalError
+				}
+				shouldSendHRR = appCookie != nil
```
ekr: Is this idiomatic go? I would have done if....

marten-seemann:
Hi, @ekr thanks for the review! Seems like I forgot a bunch of TODOs and a few debug statements. I removed those, and added a bunch of documentation about `Cookie`s, `CookieSource`s and `CookieHandler`.
The reason the tests were failing is that CircleCI apparently uses Go 1.8 if no config file is provided, and the test helper functions (`t.Helper()`) require Go 1.9. I added a basic config file.

I really like the idea of storing the negotiated parameters in the cookie as well. By doing so, I was able to remove a rather ugly hack that was necessary to make PSK mode work. Is there anything else besides the `CipherSuite` that we would want to store in the cookie?

The code for the cookie source is what we copied from Google's QUIC implementation 1.5 years ago or so. I changed the cookie source to use a nonce size of 32 byte, as you suggested, and also removed the HKDF. I'm not sure though why're suggesting using a per-message key, I thought using the same key will be ok as long as the nonce is not reused. Can you please explain?
I also made the `DefaultCookieSource` public. When using non-blocking mode, an app will have to provide its own implementation for the `CookieSource`. The `DefaultCookieSource` should work fine in easy use cases, such as running a single server (in a multi-server deployment you'll need some shared cookie source anyway, and probably also implement some key rotation mechanism).

ekr:
The issue with the nonce here is that AES-GCM fails catastrophically if you have nonce reuse. If you use a random nonce, you get a non-trivial probability of reuse with enough connections (the 50% mark is at 2^{48} but the risk is unacceptably high well below that). Unfortunately, just making the nonce bigger doesn't help because AES-GCM hashes a >96 bit nonce down to 96 bits.

If you want to use GCM, you can guarantee a unique key/nonce pair by generating a longer nonce and then using HKDF to turn the master key into a key/nonce. Because the output is longer, the chance of collision is reduced.

marten-seemann:
I just pushed a fixed. We're now creating a new key (which is HKDF derived using the nonce as a salt) for every nonce.

cookie_source.go
```diff
@@ -11,63 +11,68 @@ import (
...
+const cookieNonceSize = 32
+
+// The DefaultCookieSource is a simple implementation for the CookieSource.
+type DefaultCookieSource struct {
+	secret []byte
 }
 
-const tokenKeySize = 16
-const tokenNonceSize = 16
+var _ CookieSource = &DefaultCookieSource{}
```
ekr: I am probably not enough of a Golang wizard, but why is this neded.

cookie_source.go
```diff
@@ -11,63 +11,68 @@ import (
...
+	nonce := make([]byte, cookieNonceSize)
 	if _, err := rand.Read(nonce); err != nil {
 		return nil, err
 	}
-	return s.aead.Seal(nonce, nonce, data, nil), nil
+	aead, err := s.createAEAD(nonce)
+	if err != nil {
+		return nil, err
+	}
+	return aead.Seal(nonce, nonce, data, nil), nil
```
ekr: I don't think this is quite what you want. Rather, you want to do:

```
r := make([]byte, 32)
rand.Read(r)
key = hkdf(secret, R, []byte("mint cookie key")) // key length
nonce = hkdf(secret, R, []byte("mint cookie nonce")) // 12 bytes
```

And then you can use nonce the way you have it here.

ekr: Or, if you prefer you can just read the key and the nonce as successive values from the hkdf.

cookie_source.go
```diff
@@ -11,63 +11,68 @@ import (
...
+	key := make([]byte, 32)
 	if _, err := io.ReadFull(r, key); err != nil {
 		return nil, err
 	}
-	return key, nil
+	c, err := aes.NewCipher(key)
+	if err != nil {
+		return nil, err
+	}
+	return cipher.NewGCMWithNonceSize(c, cookieNonceSize)
```
ekr: You want NewGCM()

cookie_source_test.go
```diff
@@ -0,0 +1,27 @@
...
+		decoded, err := cs.DecodeToken(token)
+		assertNotError(t, err, "decoding the token failed")
+		assertDeepEquals(t, cookie, decoded)
+	})
+
+	t.Run("handling invalid tokens", func(t *testing.T) {
+		_, err := cs.DecodeToken([]byte("too short"))
+		assertError(t, err, "it should reject too short tokens")
+		_, err = cs.DecodeToken(append(bytes.Repeat([]byte{0}, cookieNonceSize), []byte("invalid token")...))
+		assertError(t, err, "it should reject invalid tokens")
```
ekr: It would also be good to have one with a bogus nonce, e.g., by stomping the first byte.

cookie_source.go
marten-seemann: Strictly speaking this is not needed. It makes the compiler check that `DefaultCookieSource` actually implements the `CookieSource` interface. Since it's an assignment to `_`, it will be removed during compilation.

cookie_source.go
marten-seemann: How is that supposed to work? When decoding the cookie, I don't know `r` any more, and I can't create the AEAD to open the cookie.

ekr: The cookie is r || aead.Seal(blah...)

Right now you are just using nonce == r, so the difference I am proposing is you hkdf r to make the nonce.

marten-seemann:
@ekr, I implemented the changes for the `CookieSource` as you suggested. Please have another look.

ekr:
@marten-seemann I think you forgot to push

marten-seemann:
I don't think so. I rebased and amended the commit.

ekr:
Oh, ok. Will take a look

cookie_source.go
```diff
@@ -0,0 +1,86 @@
...
+func (s *DefaultCookieSource) createAEAD(nonce []byte) (cipher.AEAD, []byte, error) {
+	h := hkdf.New(sha256.New, s.secret, nonce, []byte("mint cookie source"))
+	key := make([]byte, 32) // use a 32 byte key, in order to select AES-256
+	if _, err := io.ReadFull(h, key); err != nil {
+		return nil, nil, err
+	}
+	aeadNonce := make([]byte, 12)
+	if _, err := io.ReadFull(h, aeadNonce); err != nil {
+		return nil, nil, err
+	}
```
ekr: This looks fine to me. Assuming you didn't change anything else, I think this is good to Go. @bifurcation ??

marten-seemann: I didn't change anything else. This should be ready to merge now.

.circleci/config.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
```
bifurcation: Unless this change is necessary for this functionality, please put it in a separate PR.

marten-seemann: This was the first commit of this PR (03b32d6dfee03aaed5b3759a24fe5ef8d54cd943). I openend #147, containing just this one commit.

marten-seemann:
@bifurcation: Is there anything else you need from me to merge this PR?

cookie_source.go
```diff
@@ -0,0 +1,86 @@
+package mint
```
bifurcation: For consistency with the other files, this should be `cookie-source.go` (reserving underscores for `_test`).

conn.go
```diff
@@ -74,9 +78,16 @@ type Config struct {
 	AllowEarlyData     bool
 	// Require the client to echo a cookie.
 	RequireCookie bool
-	// If cookies are required and no CookieHandler is set, a default cookie handler is used.
-	// The default cookie handler uses 32 random bytes as a cookie.
-	CookieHandler     CookieHandler
+	// A CookieHandler can be used to set and validate a cookie.
+	// The cookie returned by the CookieHandler will be part of the cookie sent on the wire, and endoced using the CookieSource.
```
bifurcation: Nit: "encoded"

conn.go
```diff
@@ -74,9 +78,16 @@ type Config struct {
...
+	// If no CookieHandler is set, mint will always send a cookie.
+	// The CookieHandler can be used to decide on a per-connection basis, if a cookie should be sent.
+	CookieHandler CookieHandler
+	// The CookieSource is used to encrypt / decrypt cookies.
+	// It should make sure that the Cookie cannot be read and tampered with by the client.
+	// If non-blocking mode is used, and cookies are required, this field has to be set.
+	// In blocking mode, a default cookie source is used, if this is unused.
+	CookieSource      CookieSource
```
bifurcation: Nit: `CookieSource` doesn't seem like the right name for something whose job is to protect / unprotect cookie values.  `CookieProtection` or something like that?

marten-seemann:
@bifurcation: I rebased this PR on top of the current master, fixed the type, and renamed the `CookieSource` to `CookieProtector`.

bifurcation:
@marten-seemann Thanks! Merging and moving on to #145 ...

Issue #144, opened 2017-11-24
Key Share tests failing


marten-seemann:
The test `TestNewKeyShare` sometimes fails here (maybe once every 50 test runs or so): https://github.com/bifurcation/mint/blob/master/crypto_test.go#L71, because the length is 1 byte less than expected.

Furthermore, in https://github.com/bifurcation/mint/blob/master/crypto_test.go#L68, `priv` and `pub` seem to be interchanged, so maybe that's the reason this test is failing and all we need to do to fix this is swap the variables.

I don't know enough about the EC primitives here to see what the right fix is.

marten-seemann:
This also happens on the CI: https://circleci.com/gh/bifurcation/mint/263.

PR #145, opened 2017-11-25
do not automatically advance state in non-blocking mode

### LINKED ISSUE #143: In non-blocking mode, do not automatically advance state
Right now, the server state machine automatically advances in a few cases:

https://github.com/bifurcation/mint/blob/master/server-state-machine.go#L309
https://github.com/bifurcation/mint/blob/master/server-state-machine.go#L609
https://github.com/bifurcation/mint/blob/master/server-state-machine.go#L656

This is apparently causing problems for some QUIC stacks that are using mint for the TLS 1.3 handshake.  The same QUICK stacks are also the only current consumers of non-blocking mode*.  So we should not auto-advance the state machine in non-blocking mode.

* In the future, we might want a separate flag

marten-seemann:
Fixes #143. Should be merged after #141.

In non-blocking mode, `Handshake()` now returns after every state transition with `AlertNoAlert`. If `Handshake` is called, but would block while reading a new record, it returns `AlertWouldBlock`.
The state of the handshake is now returned by `Conn.GetHsState()`.

@ekr, @bifurcation: What do you think about this?

### DIFF
--- a/client-state-machine.go
+++ b/client-state-machine.go
@@ -59,12 +59,9 @@ type ClientStateStart struct {
 	helloRetryRequest *HandshakeMessage
 }
 
-func (state ClientStateStart) Next(hm *HandshakeMessage) (HandshakeState, []HandshakeAction, Alert) {
-	if hm != nil {
-		logf(logTypeHandshake, "[ClientStateStart] Unexpected non-nil message")
-		return nil, nil, AlertUnexpectedMessage
-	}
+var _ HandshakeState = &ClientStateStart{}
 
+func (state ClientStateStart) Next(hr handshakeMessageReader) (HandshakeState, []HandshakeAction, Alert) {
 	// key_shares
 	offeredDH := map[NamedGroup][]byte{}
 	ks := KeyShareExtension{
@@ -307,7 +304,13 @@ type ClientStateWaitSH struct {
 	clientHello       *HandshakeMessage
 }
 
-func (state ClientStateWaitSH) Next(hm *HandshakeMessage) (HandshakeState, []HandshakeAction, Alert) {
+var _ HandshakeState = &ClientStateWaitSH{}
+
+func (state ClientStateWaitSH) Next(hr handshakeMessageReader) (HandshakeState, []HandshakeAction, Alert) {
+	hm, alert := hr.ReadMessage()
+	if alert != AlertNoAlert {
+		return nil, nil, alert
+	}
 	if hm == nil {
 		logf(logTypeHandshake, "[ClientStateWaitSH] Unexpected nil message")
 		return nil, nil, AlertUnexpectedMessage
@@ -383,7 +386,7 @@ func (state ClientStateWaitSH) Next(hm *HandshakeMessage) (HandshakeState, []Han
 			cookie:            serverCookie.Cookie,
 			firstClientHello:  firstClientHello,
 			helloRetryRequest: hm,
-		}.Next(nil)
+		}, nil, AlertNoAlert
 
 	case *ServerHelloBody:
 		sh := body
@@ -522,15 +525,20 @@ type ClientStateWaitEE struct {
 	serverHandshakeTrafficSecret []byte
 }
 
-func (state ClientStateWaitEE) Next(hm *HandshakeMessage) (HandshakeState, []HandshakeAction, Alert) {
+var _ HandshakeState = &ClientStateWaitEE{}
+
+func (state ClientStateWaitEE) Next(hr handshakeMessageReader) (HandshakeState, []HandshakeAction, Alert) {
+	hm, alert := hr.ReadMessage()
+	if alert != AlertNoAlert {
+		return nil, nil, alert
+	}
 	if hm == nil || hm.msgType != HandshakeTypeEncryptedExtensions {
 		logf(logTypeHandshake, "[ClientStateWaitEE] Unexpected message")
 		return nil, nil, AlertUnexpectedMessage
 	}
 
 	ee := EncryptedExtensionsBody{}
-	_, err := ee.Unmarshal(hm.body)
-	if err != nil {
+	if _, err := ee.Unmarshal(hm.body); err != nil {
 		logf(logTypeHandshake, "[ClientStateWaitEE] Error decoding message: %v", err)
 		return nil, nil, AlertDecodeError
 	}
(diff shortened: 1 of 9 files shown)


marten-seemann:
Hi @ekr, this should be ready for review as well.
I'd be very happy if we could get this merged some time this week (that would be really helpful for preparing quic-go for the remote interop).

ekr:
I think @bifurcation is the first person on point for this. He also informs
me that it's failing CI...?

bifurcation:
I'm waiting to review this until #141 lands, since it's based on top of that PR.

marten-seemann:
There's nothing in this PR that is failing the CI, however, the CI is flaky because of #144.

bifurcation:
@marten-seemann Looks like this has some merge conflicts after landing #141.  Could you please resolve these?  It looks like they're mostly trivial (related to `CookieProtector`), but there are some that weren't obvious to me.

marten-seemann:
@bifurcation: Done.

client-state-machine.go
```diff
@@ -59,12 +59,13 @@ type ClientStateStart struct {
 	helloRetryRequest *HandshakeMessage
 }
 
-func (state ClientStateStart) Next(hm *HandshakeMessage) (HandshakeState, []HandshakeAction, Alert) {
-	if hm != nil {
-		logf(logTypeHandshake, "[ClientStateStart] Unexpected non-nil message")
-		return nil, nil, AlertUnexpectedMessage
-	}
+var _ HandshakeState = &ClientStateStart{}
```
bifurcation: What does this line do?  It seems like a noop.  I think you can safely remove this and the ones below it.

client-state-machine.go
```diff
@@ -59,12 +59,13 @@ type ClientStateStart struct {
...
 
+func (ClientStateStart) State() State {
```
bifurcation: Aesthetically, I would prefer to have a name for the target object, even if it's not used.

conn.go
```diff
@@ -695,35 +714,31 @@ func (c *Conn) Handshake() Alert {
 	var alert Alert
 	if c.hState == nil {
 		logf(logTypeHandshake, "%s First time through handshake (or after stateless retry), setting up", label)
-		alert = c.HandshakeSetup()
-		if alert != AlertNoAlert {
+		alert := c.HandshakeSetup()
```
bifurcation: This colon is unnecessary, since `alert` is declared above.

conn.go
```diff
@@ -695,35 +714,31 @@ func (c *Conn) Handshake() Alert {
...
+		if alert != AlertNoAlert || (c.isClient && c.config.NonBlocking) {
 			return alert
 		}
-	} else {
-		logf(logTypeHandshake, "Re-entering handshake, state=%v", c.hState)
 	}
 
+	logf(logTypeHandshake, "Re-entering handshake, state=%v", c.hState)
```
bifurcation: This log line seems wrong now that you've pulled it out of the `else` clause.  Need to change it to "(Re-)entering" or something

conn.go
```diff
@@ -741,14 +756,19 @@ func (c *Conn) Handshake() Alert {
 		c.hState = state
 		logf(logTypeHandshake, "state is now %s", c.GetHsState())
 		_, connected = state.(StateConnected)
+		if connected {
+			c.state = state.(StateConnected)
+			c.handshakeComplete = true
```
bifurcation: It seems like you could make this a little more compact:
```
if connectedState, connected := state.(StateConnected); connected {
    c.state = connectedState
    c.handshakeComplete = true
}
```

common.go
```diff
@@ -150,3 +150,66 @@ const (
 	KeyUpdateNotRequested KeyUpdateRequest = 0
 	KeyUpdateRequested    KeyUpdateRequest = 1
 )
+
+type State uint8
```
bifurcation: Adding this enum is a good idea, but should be in a separate PR.  When things are easily decoupled, I would much rather review a small PR than a large one.

conn.go
```diff
@@ -741,14 +756,19 @@ func (c *Conn) Handshake() Alert {
...
+		}
 
-		if c.config.NonBlocking && alert == AlertStatelessRetry {
-			return AlertStatelessRetry
+		if c.config.NonBlocking {
```
bifurcation: I'm a little unclear on how this affects upstream callers of `Handshake()`.  It seems like `conn.Read()`, `Listener.Accept()`, and `DialWithDialer()` all assume that `Handshake()` returning with `AlertNoAlert` means that the handshake is complete.

Those methods probably aren't suitable for use with non-blocking mode anyway (?) so maybe we just need to define a new error code that if non-blocking mode accidentally gets set, these methods will fail, but a non-blocking caller can recognize that everything's OK.

client-state-machine.go
marten-seemann: Yes, it's a noop. It's a check that the struct satisfies the interface. I like this pattern because you'll get the compile error right there if it doesn't.

common.go
marten-seemann: #151.

conn.go
marten-seemann: That doesn't work. `connected` determines when the `for` exits.

conn.go
marten-seemann: I moved the variable declaration instead, prefer to keep variables as local as possible.

conn.go
marten-seemann: Good point. Although QUIC doesn't use those, at least some of them would be useful even in non-blocking mode (e.g. you want to be able to `conn.Read()` after completing the handshake).

I think the best solution here is:
1. document that `Read` should only be called after the handshake is complete in non-blocking mode
2. not run `Handshake()`from `Read`, but 
```go
for {
  if alert := c.Handshake(); alert != AlertNoAlert {
    return alert
  }
  if _, ok := c.state.(StateConnected); ok {
     break
   }
}
```
This way, if `Read` is called before the handshake completes, at least it will fail at some point (with an `AlertWouldBlock`), but won't try to perform `Read` on an unconnected connection.

I'll add a commit to make things clearer.

marten-seemann: I just realized my idea doesn't really make sense for the listen and dial calls. Just returning an error, as you suggested, is probably the better solution. Just pushed the change, so PHAL.

marten-seemann:
@bifurcation: I think I addressed all your points.

bifurcation:
Thanks for the quick response @marten-seemann

PR #146, opened 2017-11-29
Beginning of DTLS support. This doesn't work properly yet, but it

ekr:
will handle the framing and has at least rudimentary reassembly
of out of order packets. I think it's worth getting it in to
avoid a long-lived branch.

@bifurcation this probably isn't quite ready, but I think worth a first look

### DIFF
--- a/common.go
+++ b/common.go
@@ -213,3 +213,28 @@ func (s State) String() string {
 		return fmt.Sprintf("unknown state: %d", s)
 	}
 }
+
+// Epochs for DTLS (also used for key phase labelling)
+type Epoch uint16
+
+const (
+	EpochClear           Epoch = 0
+	EpochEarlyData       Epoch = 1
+	EpochHandshakeData   Epoch = 2
+	EpochApplicationData Epoch = 3
+	EpochUpdate          Epoch = 4
+)
+
+func (e Epoch) label() string {
+	switch e {
+	case EpochClear:
+		return "clear"
+	case EpochEarlyData:
+		return "early data"
+	case EpochHandshakeData:
+		return "handshake"
+	case EpochApplicationData:
+		return "application data"
+	}
+	return "Application data (updated)"
+}
--- a/common_test.go
+++ b/common_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"reflect"
+	"runtime"
 	"testing"
 )
 
@@ -18,8 +19,16 @@ func unhex(h string) []byte {
 
 func assert(t *testing.T, test bool, msg string) {
 	t.Helper()
+	prefix := string("")
+	for i := 1; ; i++ {
+		_, file, line, ok := runtime.Caller(i)
+		if !ok {
+			break
+		}
+		prefix = fmt.Sprintf("%v: %d\n", file, line) + prefix
+	}
 	if !test {
-		t.Fatalf(msg)
+		t.Fatalf(prefix + msg)
 	}
 }
 
--- a/conn.go
+++ b/conn.go
@@ -101,6 +101,7 @@ type Config struct {
 	PSKs             PreSharedKeyCache
 	PSKModes         []PSKKeyExchangeMode
 	NonBlocking      bool
+	UseDTLS          bool
 
 	// The same config object can be shared among different connections, so it
 	// needs its own mutex
@@ -135,6 +136,7 @@ func (c *Config) Clone() *Config {
 		PSKs:             c.PSKs,
 		PSKModes:         c.PSKModes,
 		NonBlocking:      c.NonBlocking,
+		UseDTLS:          c.UseDTLS,
 	}
 }
 
@@ -252,18 +254,25 @@ type Conn struct {
 
 	readBuffer []byte
 	in, out    *RecordLayer
-	hIn, hOut  *HandshakeLayer
+	hsCtx      HandshakeContext
 
 	extHandler AppExtensionHandler
 }
 
 func NewConn(conn net.Conn, config *Config, isClient bool) *Conn {
 	c := &Conn{conn: conn, config: config, isClient: isClient}
-	c.in = NewRecordLayer(c.conn)
-	c.out = NewRecordLayer(c.conn)
-	c.hIn = NewHandshakeLayer(c.in)
-	c.hIn.nonblocking = c.config.NonBlocking
-	c.hOut = NewHandshakeLayer(c.out)
+	if !config.UseDTLS {
+		c.in = NewRecordLayerTLS(c.conn)
+		c.out = NewRecordLayerTLS(c.conn)
+		c.hsCtx.hIn = NewHandshakeLayerTLS(c.in)
+		c.hsCtx.hOut = NewHandshakeLayerTLS(c.out)
+	} else {
+		c.in = NewRecordLayerDTLS(c.conn)
+		c.out = NewRecordLayerDTLS(c.conn)
+		c.hsCtx.hIn = NewHandshakeLayerDTLS(c.in)
+		c.hsCtx.hOut = NewHandshakeLayerDTLS(c.out)
+	}
+	c.hsCtx.hIn.nonblocking = c.config.NonBlocking
 	return c
 }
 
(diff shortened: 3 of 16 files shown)


ekr:
@bifurcation: actually, I think this is in pretty good shape for what it is

common.go
```diff
@@ -150,3 +150,46 @@ const (
...
+		return "early data"
+	case EpochHandshakeData:
+		return "handshake"
+	case EpochApplicationData:
+		return "application data"
+	}
+	return "Application data (updated)"
+}
+
+func encodeUint(v uint64, size int, out []byte) []byte {
```
bifurcation: This function and `decodeUint` should move to `handshake-layer.go`.

common_test.go
```diff
@@ -17,8 +18,16 @@ func unhex(h string) []byte {
 }
 
 func assert(t *testing.T, test bool, msg string) {
+	prefix := string("")
+	for i := 1; ; i++ {
+		_, file, line, ok := runtime.Caller(i)
+		if !ok {
+			break
+		}
+		prefix = fmt.Sprintf("%v: %d\n", file, line) + prefix
```
bifurcation: Nit: Might be nicer to keep these to a single line, e.g., "file:line | file:line | file:line > msg"

conn.go
```diff
@@ -89,6 +89,7 @@ type Config struct {
 	PSKs             PreSharedKeyCache
 	PSKModes         []PSKKeyExchangeMode
 	NonBlocking      bool
+	DatagramMode     bool
```
bifurcation: Maybe it's just me, but `DatagramMode` sounds like an enum, like there might be multiple modes.  Maybe just call this option `DTLS` or `UseDTLS`?

dtls.go
```diff
@@ -0,0 +1,5 @@
+package mint
+
+const (
+	initialMtu = 1200
+)
```
bifurcation: Nit: It would be good to have a comment in this file noting that it's a stub, and what else is supposed to go here

handshake-layer.go
```diff
@@ -87,48 +104,77 @@ func (hm HandshakeMessage) ToBody() (HandshakeMessageBody, error) {
...
 func (d handshakeLayerFrameDetails) frameLen(hdr []byte) (int, error) {
 	logf(logTypeIO, "Header=%x", hdr)
-	return (int(hdr[1]) << 16) | (int(hdr[2]) << 8) | int(hdr[3]), nil
+	// The length of this fragment (as opposed to the message)
+	// is always the last three bytes for both TLS and DTLS
+	val, _ := decodeUint(hdr[len(hdr)-3:], 3)
+	return int(val), nil
+}
+
+func NewHandshakeLayerTls(r *RecordLayer) *HandshakeLayer {
```
bifurcation: Nit: `Tls` -> `TLS`

handshake-layer.go
```diff
@@ -29,26 +30,42 @@ const (
...
+	offset   uint32 // Used for DTLS
+	length   uint32
+	records  []uint64 // Used for DTLS
+	cipher   *cipherState
 }
 
 // Note: This could be done with the `syntax` module, using the simplified
 // syntax as discussed above.  However, since this is so simple, there's not
 // much benefit to doing so.
+// When datagram is set we marshal this as a whole DTLS record.
```
bifurcation: Nit: Comma after "set"

handshake-layer.go
```diff
@@ -29,26 +30,42 @@ const (
 //
 // TODO: File a spec bug
```
bifurcation: This TODO is obsolete, if you want to kill it while you've got the patient on the table.

handshake-layer.go
```diff
@@ -171,18 +216,125 @@ func (h *HandshakeLayer) sendAlert(err Alert) error {
 	return nil
 }
 
+func (h *HandshakeLayer) messageDelivered(seq uint32) {
```
bifurcation: This method name could be more descriptive.  `noteMessageDelivered` or something about queue updating?

handshake-layer.go
```diff
@@ -171,18 +216,125 @@ func (h *HandshakeLayer) sendAlert(err Alert) error {
...
+	var m *HandshakeMessage
+	for i, m = range h.queued {
+		if m.seq > seq {
+			break
+		}
+	}
+	h.queued = h.queued[i:]
+}
+
+func (h *HandshakeLayer) newFragment(hm *HandshakeMessage) (*HandshakeMessage, error) {
```
bifurcation: Similar comment about method name.  `newFragmentReceived`?

handshake-layer_test.go
```diff
@@ -83,6 +86,13 @@ func TestMessageMarshal(t *testing.T) {
 	assertByteEquals(t, out, tinyMessage)
 }
 
+func newTestHandshakeMessage(t HandshakeType, m []byte) HandshakeMessage {
```
bifurcation: Is there any particular reason this got factored out?  Planning something else for this later?

handshake-messages.go
```diff
@@ -90,10 +90,16 @@ func (ch ClientHelloBody) Truncated() ([]byte, error) {
 		return nil, fmt.Errorf("tls.clienthello.truncate: Last extension is not PSK")
 	}
 
-	chm, err := HandshakeMessageFromBody(&ch)
+	// TODO(ekr@rtfm.com): This will not work right with DTLS.
```
bifurcation: Why not?

record-layer.go
```diff
@@ -9,9 +9,10 @@ import (
 )
 
 const (
-	sequenceNumberLen = 8       // sequence number length
-	recordHeaderLen   = 5       // record header length
-	maxFragmentLen    = 1 << 14 // max number of bytes in a record
+	sequenceNumberLen   = 8       // sequence number length
+	recordHeaderLenTls  = 5       // record header length (TLS)
+	recordHeaderLenDtls = 13      // record header length (DTLS)
```
bifurcation: `Tls` -> `TLS`
`Dtls` -> `DTLS`

record-layer.go
```diff
@@ -42,72 +51,111 @@ type RecordLayer struct {
 	cachedRecord *TLSPlaintext // Last record read, cached to enable "peek"
 	cachedError  error         // Error on the last record read
 
-	ivLength int         // Length of the seq and nonce fields
-	seq      []byte      // Zero-padded sequence number
-	nonce    []byte      // Buffer for per-record nonces
-	cipher   cipher.AEAD // AEAD cipher
+	cipher   *cipherState
```
bifurcation: Do you want to just make this a `map[Epoch]*cipherState`?  Then just append to it in `Rekey()` and query it on encrypt/decrypt.  Might require special handling to remove the "special" epochs for clear, handshake, etc. once they're disallowed.

record-layer.go
```diff
@@ -42,72 +51,111 @@ type RecordLayer struct {
...
+	return d.headerLen() + maxFragmentLen
 }
 
 func (d recordLayerFrameDetails) frameLen(hdr []byte) (int, error) {
-	return (int(hdr[3]) << 8) | int(hdr[4]), nil
+	return (int(hdr[d.headerLen()-2]) << 8) | int(hdr[d.headerLen()-1]), nil
 }
 
-func NewRecordLayer(conn io.ReadWriter) *RecordLayer {
+func newCipherStateNull() *cipherState {
```
bifurcation: Is this needed?  Why not just use `nil` as the null `*cipherState`?  We already switch encryption/decryption on whether `cipher != nil`.

record-layer_test.go
```diff
@@ -25,8 +25,8 @@ func TestRekey(t *testing.T) {
 	key := unhex(keyHex)
 	iv := unhex(ivHex)
 
-	r := NewRecordLayer(bytes.NewBuffer(nil))
-	err := r.Rekey(newAESGCM, key, iv)
+	r := NewRecordLayerTls(bytes.NewBuffer(nil))
```
bifurcation: `Tls` -> `TLS`

record-layer_test.go
```diff
@@ -218,14 +217,14 @@ func TestEncryptRecord(t *testing.T) {
 	assertError(t, err, "Allowed a too-large record")
 }
 
-func TestReadWrite(t *testing.T) {
+func TestReadWriteTls(t *testing.T) {
```
bifurcation: `Tls` -> `TLS`

I'll assume from here on that you'll just `s/Tls/TLS/g; s/Dtls/DTLS/g`

common_test.go
ekr: It's pretty hard to read that way it turns out. Will do if you insist.

record-layer.go
ekr: You need it to store the sequence number

handshake-layer_test.go
ekr: Trying to remember, TBH.

handshake-messages.go
ekr: I'm trying to remember TBH. I'll remove it and I guess we'll find out :)

record-layer.go
ekr: Probably in the future, but I think later.

bifurcation:
Merging based on EKR's r+

bifurcation:
@marten-seemann I think EKR wants to use this module to serialize some other QUIC messages.

PR #150, opened 2017-12-06
Add explicit marshal/unmarshal to the TLS syntax module

bifurcation:
I think this will offer a path to implementing the `select` syntax, e.g., to make encode/decode of `HandshakeMessage` more elegant.

### DIFF
--- a/syntax/decode.go
+++ b/syntax/decode.go
@@ -16,6 +16,15 @@ func Unmarshal(data []byte, v interface{}) (int, error) {
 	return d.unmarshal(v)
 }
 
+// Unmarshaler is the interface implemented by types that can
+// unmarshal a TLS description of themselves.  Note that unlike the
+// JSON unmarshaler interface, it is not known a priori how much of
+// the input data will be consumed.  So the Unmarshaler must state
+// how much of the input data it consumed.
+type Unmarshaler interface {
+	UnmarshalTLS([]byte) (int, error)
+}
+
 // These are the options that can be specified in the struct tag.  Right now,
 // all of them apply to variable-length vectors and nothing else
 type decOpts struct {
@@ -65,8 +74,19 @@ func typeDecoder(t reflect.Type) decoderFunc {
 	return newTypeDecoder(t)
 }
 
+var (
+	unmarshalerType = reflect.TypeOf(new(Unmarshaler)).Elem()
+)
+
 func newTypeDecoder(t reflect.Type) decoderFunc {
-	// Note: Does not support Marshaler, so don't need the allowAddr argument
+	// XXX Unmarshalers are only supported by value.  That is, if Foo
+	// implements Unmarshaler, then:
+	//
+	// type A struct { f Foo }	 <-- OK
+	// type B struct { f *Foo }  <-- Not OK
+	if t.Kind() != reflect.Ptr && reflect.PtrTo(t).Implements(unmarshalerType) {
+		return unmarshalerDecoder
+	}
 
 	switch t.Kind() {
 	case reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
@@ -83,6 +103,27 @@ func newTypeDecoder(t reflect.Type) decoderFunc {
 }
 
 ///// Specific decoders below
+
+func unmarshalerDecoder(d *decodeState, v reflect.Value, opts decOpts) int {
+	um, ok := v.Interface().(Unmarshaler)
+	if !ok {
+		panic(fmt.Errorf("Non-Unarshaler passed to unmarshalerEncoder"))
+	}
+
+	read, err := um.UnmarshalTLS(d.Bytes())
+	if err != nil {
+		panic(err)
+	}
+
+	if read > d.Len() {
+		panic(fmt.Errorf("Invalid return value from UnmarshalTLS"))
+	}
+
+	d.Next(read)
+	return read
+}
+
+//////////
 
 func uintDecoder(d *decodeState, v reflect.Value, opts decOpts) int {
 	var uintLen int
--- a/syntax/decode.go
+++ b/syntax/decode.go
@@ -79,11 +79,6 @@ var (
 )
 
 func newTypeDecoder(t reflect.Type) decoderFunc {
-	// XXX Unmarshalers are only supported by value.  That is, if Foo
-	// implements Unmarshaler, then:
-	//
-	// type A struct { f Foo }	 <-- OK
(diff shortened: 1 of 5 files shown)


syntax/decode.go
```diff
@@ -16,6 +16,15 @@ func Unmarshal(data []byte, v interface{}) (int, error) {
...
 
+// Unmarshaler is the interface implemented by types that can
+// unmarshal a TLS description of themselves.  Note that unlike the
+// JSON unmarshaler interface, it is not known a priori how much of
+// the input data will be consumed.  So the Unmarshaler must state
+// how much of the input data it consumed.
+type Unmarshaler interface {
+	UnmarshalTLS([]byte) (int, error)
+}
+
```
ekr: I wonder if we should instead have this take bytes.Buffer, then you could examine Left instead. I guess that's my argument for this interface as a whole.

syntax/decode.go
```diff
@@ -65,8 +74,19 @@ func typeDecoder(t reflect.Type) decoderFunc {
 	return newTypeDecoder(t)
 }
 
+var (
+	unmarshalerType = reflect.TypeOf(new(Unmarshaler)).Elem()
+)
+
 func newTypeDecoder(t reflect.Type) decoderFunc {
-	// Note: Does not support Marshaler, so don't need the allowAddr argument
+	// XXX Unmarshalers are only supported by value.  That is, if Foo
```
ekr: I would suggest you adopt TODO(rlb@ipv.sx)...

ekr: Maybe this comment could say "can only be included in structs by value"

syntax/decode.go
```diff
@@ -84,6 +104,27 @@ func newTypeDecoder(t reflect.Type) decoderFunc {
 
 ///// Specific decoders below
 
+func unmarshalerDecoder(d *decodeState, v reflect.Value, opts decOpts) int {
+	um, ok := v.Interface().(Unmarshaler)
+	if !ok {
+		panic(fmt.Errorf("Non-Unarshaler passed to unmarshalerEncoder"))
```
ekr: Typo.

syntax/decode.go
```diff
@@ -84,6 +104,27 @@ func newTypeDecoder(t reflect.Type) decoderFunc {
...
+
+	if read > d.Len() {
+		panic(fmt.Errorf("Invalid return value from UnmarshalTLS"))
+	}
+
+	d.Next(read)
+	return read
+}
+
+//////////
```
ekr: What is this doing?

syntax/encode.go
```diff
@@ -81,6 +97,28 @@ func newTypeEncoder(t reflect.Type) encoderFunc {
...
+	if err == nil {
+		_, err = e.Write(b)
+	}
+
+	if err != nil {
+		panic(err)
+	}
+}
+
+//////////
```
ekr: Why // ?

syntax/encode_test.go
```diff
@@ -3,10 +3,50 @@ package syntax
...
 
+type CrypticString string
+
+var (
+	crypticStringMarshalCalls   = 0
+	crypticStringUnmarshalCalls = 0
+)
+
+// A CrypticString marshalls as two length octets followed by the
+// UTF-8 bytes of the string, XOR'ed with an increasing sequence
```
ekr: This appears to actually be one length octet.

syntax/encode_test.go
```diff
@@ -3,10 +3,50 @@ package syntax
...
 	"strings"
 	"testing"
 )
 
+type CrypticString string
+
+var (
+	crypticStringMarshalCalls   = 0
+	crypticStringUnmarshalCalls = 0
+)
```
ekr: These seem to be unused.

syntax/decode_test.go
```diff
@@ -152,6 +152,22 @@ func TestDecodeStruct(t *testing.T) {
 	}
 }
 
+func TestDecodeUnmarshaler(t *testing.T) {
+	var ym CrypticString
+	read, err := Unmarshal(zm, &ym)
```
ekr: These names kind of suck.

syntax/decode_test.go
```diff
@@ -152,6 +152,22 @@ func TestDecodeStruct(t *testing.T) {
...
+	if err != nil || !reflect.DeepEqual(ym, xm) || read != len(zm) {
+		t.Fatalf("struct decode failed [%v] [%v]", err, ym)
+	}
+
+	var ym2 struct {
+		Content CrypticString
+	}
+	read, err = Unmarshal(zm, &ym2)
```
ekr: Maybe a test with two values in it would be good ?

syntax/encode_test.go
```diff
@@ -149,3 +192,10 @@ func TestEncodeStruct(t *testing.T) {
 		t.Fatalf("struct encode failed [%v] [%x]", err, ys1)
 	}
 }
+
+func TestEncodeMarshaler(t *testing.T) {
+	ym, err := Marshal(xm)
+	if err != nil || !bytes.Equal(ym, zm) {
+		t.Fatalf("Marshaler encode failed [%v] [%x]", err, ym)
+	}
+}
```
ekr: Can you test the struct embedded variant?

syntax/decode.go
bifurcation: This is actually a broader issue than unmarshalers -- pointers to any type are not supported for marshal or unmarshal.  I've added a test file to demonstrate this and filed #152

syntax/decode.go
bifurcation: You mean the comment line?  It's just consistent with what happens elsewhere in the file.

syntax/encode_test.go
bifurcation: Added references to the tests, which I had forgotten.

syntax/decode.go
bifurcation: I can see why that seems appealing, but I still like the `[]byte` approach better because (1) it's consistent with `syntax.Unmarshal()`, and (2) it's a slightly simpler interface for classes to implement.

ekr: You could probably do a mixin for that, but ok

syntax/encode_test.go
```diff
@@ -28,6 +30,8 @@ func (cs CrypticString) MarshalTLS() ([]byte, error) {
 }
 
 func (cs *CrypticString) UnmarshalTLS(data []byte) (int, error) {
+	crypticStringUnmarshalCalls += 1
```
ekr: ++?

PR #151, opened 2017-12-07
introduce a way to ask mint about the state of the handshake

marten-seemann:
Should be merged after #145.

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -656,41 +656,45 @@ func TestNonblockingHandshakeAndDataFlow(t *testing.T) {
 	// Send ClientHello
 	clientAlert = client.Handshake()
 	assertEquals(t, clientAlert, AlertNoAlert)
+	assertEquals(t, client.GetHsState(), StateClientWaitSH)
 	serverAlert = server.Handshake()
 	assertEquals(t, serverAlert, AlertWouldBlock)
+	assertEquals(t, server.GetHsState(), StateServerStart)
 
 	// Release ClientHello
 	cbConn.Flush()
 
 	// Process ClientHello, send server first flight.
-	for {
+	states := []State{StateServerNegotiated, StateServerWaitFlight2, StateServerWaitFinished}
+	for _, state := range states {
 		serverAlert = server.Handshake()
-		if serverAlert == AlertWouldBlock {
-			break
-		}
 		assertEquals(t, serverAlert, AlertNoAlert)
+		assertEquals(t, server.GetHsState(), state)
 	}
+	serverAlert = server.Handshake()
+	assertEquals(t, serverAlert, AlertWouldBlock)
 
 	clientAlert = client.Handshake()
 	assertEquals(t, clientAlert, AlertWouldBlock)
 
 	// Release server first flight
 	sbConn.Flush()
-	clientAlert = client.Handshake()
-	assertEquals(t, clientAlert, AlertNoAlert)
+	states = []State{StateClientWaitEE, StateClientWaitCertCR, StateClientWaitCV, StateClientWaitFinished, StateClientConnected}
+	for _, state := range states {
+		clientAlert = client.Handshake()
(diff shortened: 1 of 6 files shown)


marten-seemann:
@bifurcation: I just rebased this PR, should be ready for merging now.

conn_test.go
```diff
@@ -656,41 +656,56 @@ func TestNonblockingHandshakeAndDataFlow(t *testing.T) {
...
 
 	// Process ClientHello, send server first flight.
-	for {
-		serverAlert = server.Handshake()
-		if serverAlert == AlertWouldBlock {
-			break
-		}
-		assertEquals(t, serverAlert, AlertNoAlert)
-	}
+	serverAlert = server.Handshake()
```
bifurcation: Might be more elegant to do this as a list + for-loop.

marten-seemann: @bifurcation: Done.

ekr:
@marten-seemann: OK, so this is kind of unfortunate, I relied on being able to run the handshake and have all its states run to completion. Now, it seems like I have to loop around the handshake and wait for the state to be at a specific point where it's waiting for input. Is that what you do?

marten-seemann:
Yes, that's correct, you'll need to call `State()` after every step and check if it is `mint.State{Client,Server}Connected`. On the other hand, I was able to get rid of using `mint.AlertWouldBlock`, so that made the code a bit easier.

ekr:
That doesn't work right for what I need, because I don't want to know just when it's *connected* but rather when it's done everything that it can do without more data. So now I'm going to need a new way to ask "do you need more data, or do you just need me to call handshake() again". Ugh.

ekr:
I'm now regretting that I didn't review this myself. Can you explain why you need this semantic? To the best of my knowledge, no other TLS stack does this.

marten-seemann:
I didn't look at how you're doing things in minq (point me to the file, then I can have a look), but you still can use `AlertWouldBlock` if you rely on that. I just found it a nice property that I don't need it any more in quic-go.

In quic-go we have the [session](https://github.com/lucas-clemente/quic-go/blob/master/session.go), which is a pretty complex object, handling s loss recovery, flow control, congestion, various timers and much more. Using this API, I can defer allocating this much state until *after* I received the ClientHello containing a valid cookie, but *before* sending the certificates (so certificates will already be covered by the normal QUIC retransmission mechanism).

ekr:
The code is in https://github.com/ekr/minq/blob/master/tls.go#L97

The problem is that you are returning AlertNoAlert after processing CH but
before sending any data.
It looks like I can loop around if alert is AlertNoAlert *and* state !=
Connected, except that in the case of StatelessRetry .hs is nil and so
GetHsState() segvs.

marten-seemann:
You want some special treatment of StatelessRetry, don't you? The whole point of performing a stateless retry is that you can discard all your state.

I don't understand your last sentence. When is `GetHsState` nil?

ekr:
> You want some special treatment of StatelessRetry, don't you?

No.

> The whole point of performing a stateless retry is that you can discard all your state.

You're conflating HRR with StatelessRetry. HRR doesn't *need* to be used with the server discarding state, but, yes, it allows that.  Previously, Mint allowed you to perform the whole CH-HRR-CH cycle on the same mint.Connection, and that's how Minq uses it. I may make Minq purely stateless in the future, but right now, I don't. Unfortunately, after your patch, if you call GetHsState() on a mint.Connection after it has sent HRR, you get a null pointer deref.

marten-seemann:
I see. I assumed that if a mint user goes through the trouble of using the non-blocking mode, he would only do so to use the stateless retry. 

If you don't want to make minq stateless at this moment, you could just recreate the `mint.Conn` when an HRR is sent. I can't reproduce the nil pointer dereference (I get `StateServerStart` when calling `State()` after mint returned the `AlertStatelessRetry`), so I'm not sure what to do about that. Maybe there's an easy fix for that?

PR #154, opened 2017-12-07
Add support for the QUIC varint encoding

bifurcation:
This PR adds a struct tag `varint` that can be included in the `tls` annotations, without any argument.  It causes the encoder to automatically select the minimal size varint.  The decoder will decode any varint that is smaller than the allowed size for the type of the field.

(I thought about allowing the annotation to specify the length of the varint, and it's possible, but it's a little hairy.  It works fine in the encode direction, but in the decode direction, you need an extra check to make sure you don't overflow.  And that seems kind of awkward for the caller to grok.  And in any case, if you need to specify the size of a varint, it's not very "var", is it?)

I also did some related cleanup due to discovering `reflect.Type.Size()` and `bytes.Buffer.Next()`.  The former allowed some `switch` statements to be deleted; the latter shortens calls to `Read()` by a few lines.


marten-seemann:
Why does this need to be in mint? This seems like something that belongs into a QUIC implementation instead of a TLS implementation. The QUIC Transport Parameters don't even use varint encoding, so where would this code actually be used?

syntax/decode.go
```diff
@@ -121,34 +122,52 @@ func unmarshalerDecoder(d *decodeState, v reflect.Value, opts decOpts) int {
...
 
+	return setUintFromBuffer(v, buf)
+}
+
+func varintDecoder(d *decodeState, v reflect.Value, opts decOpts) int {
+	// Read the first octet and decide the size of the presented varint
+	first := d.Next(1)
+	if len(first) != 1 {
+		panic(fmt.Errorf("Insufficient data to read varint length"))
+	}
```
ekr: Just to confirm, this panic gets caught above, right?

syntax/decode.go
```diff
@@ -121,34 +122,52 @@ func unmarshalerDecoder(d *decodeState, v reflect.Value, opts decOpts) int {
...
+		panic(fmt.Errorf("Uint too small to fit varint: %d < %d"))
+	}
+
+	rest := d.Next(varintLen - 1)
+	if len(rest) != varintLen-1 {
+		panic(fmt.Errorf("Insufficient data to read varint"))
+	}
+
+	buf := append(first, rest...)
+	buf[0] ^= byte(twoBits) << 6
```
ekr: I think I would mask off the lower bits instead. I.e.,
```
buf[0] = byte(twoBits) & (0x63)

File: syntax/encode.go [hunk 3]
```diff
@@ -116,18 +117,45 @@ func marshalerEncoder(e *encodeState, v reflect.Value, opts encOpts) {
...
+	for _, len := range []uint{1, 2, 4, 8} {
+		if u < (uint64(1) << (8*len - 2)) {
+			varintLen = int(len)
+			break
+		}
+	}
+
+	if uintLen < varintLen {
+		panic(fmt.Errorf("Uint too small to fit varint: %d < %d"))
+	}
```
ekr: How can this happen?

syntax/encode.go
```diff
@@ -116,18 +117,45 @@ func marshalerEncoder(e *encodeState, v reflect.Value, opts encOpts) {
...
+
+	twoBits := map[int]uint64{1: 0x00, 2: 0x01, 4: 0x02, 8: 0x03}[varintLen]
+	shift := uint(8*varintLen - 2)
+	writeUint(e, u^(twoBits<<shift), varintLen)
```
ekr: Why are you using XOR here rather than |? I realize they are the same in this context,but it seems confusing

syntax/encode.go
```diff
@@ -116,18 +117,45 @@ func marshalerEncoder(e *encodeState, v reflect.Value, opts encOpts) {
...
+}
+
+func writeUint(e *encodeState, u uint64, len int) {
+	data := make([]byte, len)
+	for i := 0; i < len; i += 1 {
+		data[i] = byte(u >> uint(8*(len-i-1)))
```
ekr: It's probably the same, but my usual idiom is to shift |u| rather than re-shifting the whole way each time

syntax/decode.go
bifurcation: Yep.  This follows the general pattern (from `encoding/json`) where there's a top-level `recover()` and everything else uses `panic()`.

syntax/encode.go
bifurcation: It can't.  Vestige of my first draft, which let the caller specify the varint length.

syntax/encode.go
bifurcation: No reason.  Happy to switch if you think it's more intuitive.

syntax/encode.go
bifurcation: I don't think that works, since we're reading from high-order bits to low-order.  If were the other way around we could do something like `data[i] = byte(u); u >>=8`, but if you pre-shift going high-to-low, you erase the stuff you'll need later.

bifurcation:
Merging based on EKR's prior r+

@marten-seemann I think EKR was thinking about using this module to encode/decode some other QUIC messages.

PR #155, opened 2017-12-11
do not need to select cert if there is only one cert set

Neilpang:


bifurcation:
@Neilpang @marten-seemann CI is failing on this.  Is this PR still needed?

Neilpang:
@bifurcation  Yes, I still think it necessary.

marten-seemann:
I think we need something along these lines. In the standard library, a TLS server would use the first certificate if only one is specified **and** the `tls.Config.GetCertificate` callback is not set.

I don't yet understand why mint checks for the signature algorithms when selecting a certificate (or rather, I don't understand why the standard library doesn't do that here: https://golang.org/src/crypto/tls/common.go#L712). Is this due to a change in TLS 1.3?

bifurcation:
Actually, on further reflection, I disagree with the proposal here.  If the server doesn't have a certificate that matches the name the client is asking for, then it should fail rather than returning a default certificate.  We have seen in other cases that [default certificates are bad for security](https://www.ietf.org/mail-archive/web/acme/current/msg00524.html).

@marten-seemann - I can't speak to why the standard library omits that check :)  The reason it's needed is that the signature algorithms in the ClientHello are supposed to represent everything the client supports, so if the certificate you pick doesn't match one of them, then the client won't be able to verify the handshake.  As with the above point, the server shouldn't give the client something that it doesn't expect the client to be able to validate.

PR #156, opened 2017-12-19
Draft 22

ekr:


ekr:
@bifurcation This doesn't implement the compat mode stuff and will fail on ChangeCipherSpec. I plan to submit that as a separate PR. LMK if you want to land those together.

common.go
```diff
@@ -22,6 +21,9 @@ const (
 	RecordTypeApplicationData RecordType = 23
 )
 
+const tls10Version uint16 = 0x0301
+const tls12Version uint16 = 0x0303
```
bifurcation: Suggest changing the `var` block above to `const` and just adding these up there.

handshake-messages.go
```diff
@@ -28,9 +28,10 @@ type ClientHelloBody struct {
 	// Omitted: clientVersion
 	// Omitted: legacySessionID
```
bifurcation: Please remove this comment line, since we now have `LegacySessionID`

handshake-messages.go
```diff
@@ -150,10 +152,12 @@ func (hrr *HelloRetryRequestBody) Unmarshal(data []byte) (int, error) {
 //     Extension extensions<0..2^16-1>;
 // } ServerHello;
 type ServerHelloBody struct {
-	Version     uint16
-	Random      [32]byte
-	CipherSuite CipherSuite
-	Extensions  ExtensionList `tls:"head=2"`
+	Version                 uint16
+	Random                  [32]byte
+	LegacySessionID         []byte `tls:"head=1,max=32"`
```
bifurcation: Please update the comment above with the new ServerHello structure.

PR #160, opened 2018-01-09
privatize handshake states

marten-seemann:
This PR privatizes all handshake states, thereby reducing our [public API](https://godoc.org/github.com/bifurcation/mint) by roughly 20%.

In general, I think we should only expose objects and functions that are actually needed to interact with mint. There is **a lot** more that we can privatize, so this PR is just a small step in this direction. If we agree that this is the path we want to go, I'll be happy to contribute more PRs.

@bifurcation, @ekr: What do you think?


ekr:
I agree we should privatize as much of the API as possible and this seems like a good start. @bifurcation:?

marten-seemann:
@bifurcation?

bifurcation:
I'm good with this in principle, but now there are conflicts.  @marten-seemann could you rebase?

marten-seemann:
@bifurcation: Rebased it.

Issue #161, opened 2018-01-09
certificate validation not performed by default


marten-seemann:
In order to verify the certificate sent by the peer, it is necessary to set the `mint.Config.AuthCertificate` callback. This is insecure and unexpected (the TLS implementation in the standard library verifies the certificate by default), and undocumented as well.

What is the reason that mint behaves this way? Can we fix this?

marten-seemann:
I just noticed that even by setting `mint.Config.AuthCertificate`, certificate verification doesn't work, because

1. `AuthCertificate` is not copied from the `Config` to the `Caps` in https://github.com/bifurcation/mint/blob/master/conn.go#L641-L654.
2. `AuthCertificate` is not copied from the `Caps` to  the `ClientStateWaitCertCR` in https://github.com/bifurcation/mint/blob/master/client-state-machine.go#L694.

This issue should be classified as **critical**. mint accepts *every* certificate.

@ekr @bifurcation

bifurcation:
Thanks for flagging this.  I will take a stab at it tonight/tomorrow.

PR #162, opened 2018-01-11
expose the peer certificates in the connection state

marten-seemann:
The `ConnectionState.PeerCertificates` is now filled with the certificate chain that the peer sent. This works for both the server as well as the client.

Note that this PR is unrelated to #161. The certificate chain exposed via `ConnectionState.PeerCertificates` is just what the peer sent.

@ekr: This was marked as a TODO for you in the code, so I guess you'll have the most context to review this PR, right?


marten-seemann:
@ekr, @bifurcation: Any thoughts on this PR?

client-state-machine.go
```diff
@@ -821,6 +822,10 @@ func (state ClientStateWaitCV) Next(hr handshakeMessageReader) (HandshakeState,
 	} else {
 		logf(logTypeHandshake, "[ClientStateWaitCV] WARNING: No verification of server certificate")
 	}
+	serverCertChain := make([]*x509.Certificate, len(state.serverCertificate.CertificateList))
+	for i, certEntry := range state.serverCertificate.CertificateList {
+		serverCertChain[i] = certEntry.CertData
+	}
```
ekr: Does the certificate authenticator do path construction?

conn_test.go
```diff
@@ -109,75 +113,48 @@ func newBufferedConn(p net.Conn) *bufferedConn {
...
+	return key, cert, err
+}
+
+func init() {
+	var err error
+	serverKey, serverCert, err = genCerts()
+	if err != nil {
+		panic(err)
+	}
+	clientKey, clientCert, err = genCerts()
```
ekr: You probably don't want this to use the server name.

ekr: Can you use newSelfSigned()

client-state-machine.go
marten-seemann: I'm not sure I understand your point. Certificate verification is broken anyway (#161).

ekr:
On Fri, Jan 26, 2018 at 3:08 PM, Marten Seemann <notifications@github.com>
wrote:

> *@marten-seemann* commented on this pull request.
> ------------------------------
>
> In client-state-machine.go
> <https://github.com/bifurcation/mint/pull/162#discussion_r164243708>:
>
> > @@ -821,6 +822,10 @@ func (state ClientStateWaitCV) Next(hr handshakeMessageReader) (HandshakeState,
>  	} else {
>  		logf(logTypeHandshake, "[ClientStateWaitCV] WARNING: No verification of server certificate")
>  	}
> +	serverCertChain := make([]*x509.Certificate, len(state.serverCertificate.CertificateList))
> +	for i, certEntry := range state.serverCertificate.CertificateList {
> +		serverCertChain[i] = certEntry.CertData
> +	}
>
> I'm not sure I understand your point. Certificate verification is broken
> anyway (#161 <https://github.com/bifurcation/mint/issues/161>).
>

Yes, but eventually it will be fixed, and so it would be good to have the
semantics be correct when that happens.

-Ekr

conn_test.go
marten-seemann: Done.

marten-seemann:
I don't think we can do this change without fixing #161. I'll submit a PR for #161 very soon, and I hope we can move forward with this quickly. This is a blocking issue for me currently, since I need a way to authenticate the peer.

bifurcation:
Closing in favor of #173

PR #163, opened 2018-01-12
Require AuthCertificate when certificates are used

### LINKED ISSUE #161: certificate validation not performed by default

bifurcation:
Fixes #161

### DIFF
--- a/client-state-machine.go
+++ b/client-state-machine.go
@@ -614,7 +614,7 @@ func (state ClientStateWaitEE) Next(hr handshakeMessageReader) (HandshakeState, 
 
 	logf(logTypeHandshake, "[ClientStateWaitEE] -> [ClientStateWaitCertCR]")
 	nextState := ClientStateWaitCertCR{
-		AuthCertificate:              state.AuthCertificate,
+		AuthCertificate:              state.Caps.AuthCertificate,
 		Params:                       state.Params,
 		hsCtx:                        state.hsCtx,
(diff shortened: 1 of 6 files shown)


bifurcation:
@marten-seemann For some reason, the UI won't let me request review from you.  WDYT?

marten-seemann:
I guess I'm not allowed to review because I'm only a "Contributor" and not a "Collaborator", in Github terms.

Thanks for the quick fix. I think, in general there are two ways to fix #161:
1. require `AuthCertificate` to be set, when using certificates
2. don't use `AuthCertificate`, but always do certificate verification

Your PR does option 1, and I think it's a good thing to throw and error when no verification is performed.

I'm wondering though if we need an `AuthCertificate` at all. The Go standard library doesn't have one, but always performs certificate verification (unless `tls.Config.InsecureSkipVerify` is set in the `tls.Config`). If an wishes to additional verification, it can set the `tls.Config.VerifyPeerCertificate`, which is performed after the standard certificate verification. I'd be happy to contribute some code for that, if we decide to do something analogous to this.

bifurcation:
Closing in favor of @marten-seemann's recent PRs.

PR #165, opened 2018-01-20
Add support for encoding and decoding structs with varint and 0-length

ekr:
headers.

### DIFF
--- a/syntax/README.md
+++ b/syntax/README.md
@@ -72,3 +72,16 @@ The available annotations right now are all related to vectors:
   fragment[TLSPlaintext.length]`.  Note, however, that in cases where the length
   immediately preceds the array, these can be reframed as vectors with
   appropriate sizes.
+
+
+QUIC Extensions Syntax
+======================
+syntax also supports some minor extensions to allow implementing QUIC.
+
+* The `varint` annotation describes a QUIC-style varint
+* `head=0` means no header, i.e., the bytes are encoded directly on the wire.
+  On reading, the decoder will consume all available data.
+* `head=255` means to encode the header as a varint
+
+
+
--- a/syntax/README.md
+++ b/syntax/README.md
@@ -79,9 +79,6 @@ QUIC Extensions Syntax
 syntax also supports some minor extensions to allow implementing QUIC.
 
 * The `varint` annotation describes a QUIC-style varint
-* `head=0` means no header, i.e., the bytes are encoded directly on the wire.
+* `head=none` means no header, i.e., the bytes are encoded directly on the wire.
   On reading, the decoder will consume all available data.
-* `head=255` means to encode the header as a varint
-
-
-
+* `head=varint` means to encode the header as a varint
--- a/syntax/decode.go
+++ b/syntax/decode.go
@@ -138,19 +138,27 @@ func uintDecoder(d *decodeState, v reflect.Value, opts decOpts) int {
 }
 
 func varintDecoder(d *decodeState, v reflect.Value, opts decOpts) int {
+	l, val := readVarint(d)
+
+	uintLen := int(v.Elem().Type().Size())
+	if uintLen < l {
+		panic(fmt.Errorf("Uint too small to fit varint: %d < %d", uintLen, l))
+	}
+
+	v.Elem().SetUint(val)
+
+	return l
+}
+
+func readVarint(d *decodeState) (int, uint64) {
 	// Read the first octet and decide the size of the presented varint
 	first := d.Next(1)
 	if len(first) != 1 {
 		panic(fmt.Errorf("Insufficient data to read varint length"))
 	}
 
-	uintLen := int(v.Elem().Type().Size())
 	twoBits := uint(first[0] >> 6)
 	varintLen := 1 << twoBits
-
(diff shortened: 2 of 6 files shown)


marten-seemann:
Why does this belong in mint? Isn’t this something that belongs into a QUIC implementation?

ekr:
Syntax isn't exactly part of Mint, it's a submodule which you can use generically in any protocol with TLS-style encoding. And it's a lot more convenient to be able to intermix varint structures with non-varint

syntax/README.md
```diff
@@ -72,3 +72,16 @@ The available annotations right now are all related to vectors:
...
+QUIC Extensions Syntax
+======================
+syntax also supports some minor extensions to allow implementing QUIC.
+
+* The `varint` annotation describes a QUIC-style varint
+* `head=0` means no header, i.e., the bytes are encoded directly on the wire.
+  On reading, the decoder will consume all available data.
+* `head=255` means to encode the header as a varint
+
+
```
bifurcation: Nit: Extra whitespace

syntax/decode.go
```diff
@@ -138,37 +138,50 @@ func uintDecoder(d *decodeState, v reflect.Value, opts decOpts) int {
 }
 
 func varintDecoder(d *decodeState, v reflect.Value, opts decOpts) int {
+	l, val := readVarint(d)
+
+	uintLen := int(v.Elem().Type().Size())
+	if uintLen < l {
```
bifurcation: Should this test for `==` instead of `<`?  Might be more conservative to pass the desired length into `readVarint` and have it fail if the varint isn't the right size.

syntax/decode.go
```diff
@@ -200,36 +213,53 @@ type sliceDecoder struct {
...
-
-	if opts.max > 0 && length > opts.max {
-		panic(fmt.Errorf("Length of vector exceeds declared max"))
-	}
-	if length < opts.min {
-		panic(fmt.Errorf("Length of vector below declared min"))
-	}
+	// If head == 0, then read everything from the buffer
+	if opts.head == 0 {
+		for {
```
bifurcation: You can just call `d.Bytes()` to get all of the remaining unread data.

syntax/README.md
```diff
@@ -72,3 +72,16 @@ The available annotations right now are all related to vectors:
...
   immediately preceds the array, these can be reframed as vectors with
   appropriate sizes.
+
+
+QUIC Extensions Syntax
+======================
+syntax also supports some minor extensions to allow implementing QUIC.
+
+* The `varint` annotation describes a QUIC-style varint
+* `head=0` means no header, i.e., the bytes are encoded directly on the wire.
```
bifurcation: This seems brittle to me -- someone forgets to add a `head` tag things go weird instead of just failing.  Instead, let's define some tokens to go here that indicate what we want, e.g., `head=none`, `head=varint`.  You can map those to const integer values in `tags.go:parseTag()`, as is done for the [`varint` tag itself](https://github.com/ekr/mint/blob/a20fcf6aef8132bc4042baa6a8c7773b0dfe4bc0/syntax/tags.go#L21)

bifurcation: And in case it wasn't clear, once you do this, define consts in `tags.go` for the relevant values:

```
const (
  headValueNoHead = 255
  headValueVarint = 254
)
```

That will make the encode/decode functions more readable.

syntax/decode_test.go
```diff
@@ -153,6 +146,46 @@ func TestDecodeSlice(t *testing.T) {
 	}
 }
 
+func TestDecodeSliceZeroHead(t *testing.T) {
+	type inner struct {
+		X []byte
+	}
+
+	encoded := bytes.Repeat([]byte{0xA0}, 64)
```
bifurcation: Please follow the `x/y/z` convention that the other test cases follow.  Suggest `xv0` and `xvV`.

syntax/encode_test.go
```diff
@@ -237,6 +232,37 @@ func TestEncodeStruct(t *testing.T) {
 	}
 }
 
+func TestEncodeSliceZeroHead(t *testing.T) {
+	type inner struct {
+		x []byte `tls:"head=0"`
+	}
+
+	s := bytes.Repeat([]byte{0xA0}, 10)
```
bifurcation: Same comment here about `x/y/z` naming convention.

syntax/decode.go
```diff
@@ -200,36 +213,56 @@ type sliceDecoder struct {
...
 
 func (sd *sliceDecoder) decode(d *decodeState, v reflect.Value, opts decOpts) int {
+	var length uint64
+	var read int
+	var data []byte
+
 	if opts.head == 0 {
 		panic(fmt.Errorf("Cannot decode a slice without a header length"))
 	}
+	// If head == 0, then read everything from the buffer
```
bifurcation: Nit: This comment is now incorrect:
```
// If the caller indicates there is no header, then read everything from the buffer
```

syntax/decode.go
bifurcation: Comment withdrawn.  On further reflection, this is wrong.

PR #166, opened 2018-01-28
Refactor syntax tests

bifurcation:
This PR refactors the tests in the `syntax` module to be much cleaner.  In particular, it gets rid of the `x/y/z` notation that had been used for known-answer values.  The new structure is as follows:

* Successful encode/decoded tests are done automatically in `success_test.go`, simply by providing a value and its correct encoding
* Encode/decode error cases are tested in `encode_test.go` and `decode_test.go`
* All tests methods follow the "list of test cases followed by common test logic" pattern

With this refactor, it was easy to spot a few uncovered spots and add tests for them.  Coverage in this module is no ~95%, vs. ~85% before.

Also, I fixed a minor comment bug left over from #165


bifurcation:
Also, the diff here is not going to be that useful for review.  Suggest just reviewing the following files directly:

* `success_tests.go`
* `encode_tests.go`
* `decode_tests.go`

syntax/success_test.go
```diff
@@ -0,0 +1,234 @@
...
+			value: struct {
+				V uint64 `tls:"varint"`
+			}{V: 0x3FFFFFFFFFFFFFFF},
+			encoding: unhex("FFFFFFFFFFFFFFFF"),
+		},
+
+		// Arrays
+		"array": {
+			value:    [5]uint16{0x1111, 0x2222, 0x3333, 0x4444, 0x5555},
+			encoding: unhex("11112222333344445555"),
```
ekr: A test where the integers had different top and bottom bytes might be more intersting

syntax/success_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+func hexBuffer(size int) string {
+	return strings.Repeat("A0", size)
+}
+
+type CrypticString string
+
+var (
+	crypticStringMarshalCalls   = 0
+	crypticStringUnmarshalCalls = 0
```
ekr: Do you check these?

bifurcation: No, and I don't think there's much value to checking, and it's hard to do in the current framework.  So I've deleted them.

PR #167, opened 2018-01-29
don't generate a self-signed certificate by default

marten-seemann:
This clearly doesn't belong a general purpose TLS implementation.


ekr:
This commit broke not only Minq but also broke the auto-generating functions in the server programs.

What needed to happen here was that newSelfSigned became a public API that other programs could use, no that it was made unavailable and moved into tests.

ekr:
Here, for instance, is what main.go does:

2018/02/18 14:54:43 server: listen: tls: neither Certificates nor GetCertificate set in Config
exit status 1
14:54:43|~/dev/go/src/github.com/bifurcation/mint (master)$

PR #168, opened 2018-01-29
privatize initializing the Config

marten-seemann:


marten-seemann:
This needs some more work, and should cover `ValidForClient` and `ValidForServer` as well.

PR #169, opened 2018-01-29
set the extension handler in the Config, not on the Conn

marten-seemann:
This is more consistent with how the `Config` works (e.g. for the `Config.CookieProtector` and the `Config.CookieHandler`. Furthermore, this refactoring was necessary for implementing the certificate verification in #170.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -88,7 +88,9 @@ type Config struct {
 	// It should make sure that the Cookie cannot be read and tampered with by the client.
 	// If non-blocking mode is used, and cookies are required, this field has to be set.
 	// In blocking mode, a default cookie protector is used, if this is unused.
-	CookieProtector   CookieProtector
+	CookieProtector CookieProtector
+	// The ExtensionHandler is used to add custom extensions.
+	ExtensionHandler  AppExtensionHandler
 	RequireClientAuth bool
 
 	// Shared fields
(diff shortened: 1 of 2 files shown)


ekr:
This actually broke Minq, which depended on this API being the way it was.

I'd like to avoid situations where we have unexpected bustage like this and make sure we have an opportunity to detect them and discuss.

@bifurcation: Open to various possibilities but a simple one would be to add Minq to the Mint CI suite.

marten-seemann:
@ekr: I think the recommended solution for this in Go is vendoring your dependencies. I’m doing this in quic-go as well.
I don’t think we should add any third-party projects depending on mint to its CI build.

ekr:
That's a reasonable solution when you have two very independent projects and the dependency is good about keeping a stable API. But the problem here is that we are simultaneously improving Mint and working on the dependent projects (Minq and quic-go) and so changes are slipping in unnoticed that break one of the dependencies (as well as breaking the demo programs, as I just finished fixing). 
So, what I want is some mechanism for detecting when that happens so I don't have a big hassle whenever I want to move forward in Mint. Running Minq in CI is one option, never making breaking changes to the API without much clearer signoff is another. But the current thing isn't working

marten-seemann:
Unstable APIs are the primary reason for vendoring dependencies, so I think our both use cases exactly match this. I know that API changes suck (I didn't vendor mint in the beginning, and the constant CI fails drove me crazy). I wish Go had a better dependency management system, but vendoring dependencies is what the Go team went with...

I don't think we should add minq (or quic-go) in the CI here. We will probably need to do shome breaking API changes to mint in the future, and I don't want to be responsible for the corresponding minq fixes, and I don't think you'd want to be responsible for the corresponding quic-go changes. 

We can make it clearer our PRs here if they introduces a breaking API change, however, considering our current unnecessarily big API I suspect that this will apply to probably every PR that does more than just refactoring.

PR #170, opened 2018-01-29
implement certificate verification

### LINKED ISSUE #161: certificate validation not performed by default

marten-seemann:
Fixes #161.
Depends on #169.

The code for the cert verification was taken from the Go standard library: https://golang.org/src/crypto/tls/handshake_client.go#L283.


bifurcation:
Not crazy about pulling more X.509 logic into this library; typically like to keep the two separate.  But I agree that it increases usability a bit, and the footprint doesn't seem that big, so I'll allow it.

PR #174, opened 2018-02-18
Restore self signing utils

ekr:
This PR cleans up fallout from PR#167 and PR#170. Specifically:

1. Exposes a public MakeNewSelfSignedCert API because need that several places.
2. Modifies the server programs so that they explicitly call the new API (otherwise they don't work).
3. Adds a flag to the clients to skip cert verification.

In future, it would help if we were more careful

### DIFF
--- a/bin/mint-server-https/main.go
+++ b/bin/mint-server-https/main.go
@@ -153,6 +153,12 @@ func main() {
 		log.Fatalf("Error: %v", err)
 	}
 
+	if certFile == "" && keyFile == "" {
+		var cert *x509.Certificate
+		priv, cert, err = mint.MakeNewSelfSignedCert(serverName, mint.RSA_PKCS1_SHA256)
+		certChain = []*x509.Certificate{cert}
+
+	}
 	// Load response file
 	if responseFile != "" {
 		log.Printf("Loading response file: %v", responseFile)
@@ -177,15 +183,16 @@ func main() {
 
 	config.SendSessionTickets = sendTickets
 
-	if certChain != nil && priv != nil {
+	if certFile != "" && keyFile != "" {
 		log.Printf("Loading cert: %v key: %v", certFile, keyFile)
-		config.Certificates = []*mint.Certificate{
-			{
-				Chain:      certChain,
-				PrivateKey: priv,
-			},
-		}
 	}
+	config.Certificates = []*mint.Certificate{
+		{
+			Chain:      certChain,
+			PrivateKey: priv,
+		},
+	}
+
 	config.Init(false)
 
 	service := "0.0.0.0:" + port
--- a/bin/mint-server-https/main.go
+++ b/bin/mint-server-https/main.go
@@ -25,6 +25,7 @@ var (
 	responseFile string
 	h2           bool
 	sendTickets  bool
+	genCert      bool
 )
 
 type responder []byte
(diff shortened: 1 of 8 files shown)


crypto.go
```diff
@@ -616,3 +618,50 @@ func makeTrafficKeys(params CipherSuiteParams, secret []byte) keySet {
 		iv:     HkdfExpandLabel(params.Hash, secret, "iv", []byte{}, params.IvLen),
 	}
 }
+
+func MakeNewSelfSignedCert(name string, alg SignatureScheme) (crypto.Signer, *x509.Certificate, error) {
```
marten-seemann: I don't think generating a self-signed certificate should be part of a TLS library.

ekr: Yeah, I obviously don't agree with this. We have two test programs that rely on it and I also use it in Minq. So, I really don't think it's sensible to inconvenience people out of a sense of API purity.

ekr:
On the topic of CI, there's no reason we can't write a shell script to test these programs. It just needs doing.

bin/mint-server-https/main.go
```diff
@@ -153,6 +153,12 @@ func main() {
 		log.Fatalf("Error: %v", err)
 	}
 
+	if certFile == "" && keyFile == "" {
+		var cert *x509.Certificate
+		priv, cert, err = mint.MakeNewSelfSignedCert(serverName, mint.RSA_PKCS1_SHA256)
+		certChain = []*x509.Certificate{cert}
+
```
bifurcation: Nit: Extra blank line

bin/mint-server-https/main.go
```diff
@@ -153,6 +153,12 @@ func main() {
 		log.Fatalf("Error: %v", err)
 	}
 
+	if certFile == "" && keyFile == "" {
```
bifurcation: Let's explicitly signal this with a flag.

bifurcation: ... and obviously fail if these files are missing and the flag isn't set.

bin/mint-server-https/main.go
```diff
@@ -177,15 +183,16 @@ func main() {
 
 	config.SendSessionTickets = sendTickets
 
-	if certChain != nil && priv != nil {
+	if certFile != "" && keyFile != "" {
```
bifurcation: Seems like this should get consolidated with the above:
```
switch {
case selfSign: /* generate a key/cert */
case certFile != "" && keyFile != "": /* load a key/cert */
default: /* fail */
}
```

conn_test.go
```diff
@@ -180,19 +145,11 @@ const (
...
-	serverKey, err := rsa.GenerateKey(rand.Reader, 2048)
-	if err != nil {
-		panic(err)
-	}
-	serverCert, err = newSelfSigned(serverName, RSA_PKCS1_SHA256, serverKey)
-	if err != nil {
-		panic(err)
-	}
-	clientKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	serverKey, serverCert, err = MakeNewSelfSignedCert(serverName, RSA_PKCS1_SHA256)
```
bifurcation: Y U NO ECDSA

crypto.go
bifurcation: +1 to EKR.  We've already crossed the bridge of importing PKI gunk by following the `crypto/tls` approach of having a built-in validation system.  This is a sensible utility to offer folks.

conn_test.go
ekr: I can change to ECDSA. The existing code had RSA, so I was just trying to avoid changing the behavior

ekr: @bifurcation: just LMK

PR #177, opened 2018-02-23
Enhance dtls2

ekr:
This is really enormous, and as noted in the comments, not quite done, but I also want to get it landed before we have more rebase issues.

@bifurcation: LMK if you want me to file issues for the known deficiencies

### DIFF
--- a/client-state-machine.go
+++ b/client-state-machine.go
@@ -175,7 +175,7 @@ func (state clientStateStart) Next(hr handshakeMessageReader) (HandshakeState, [
 		// TODO(ekr@rtfm.com): Check that the ticket can be used for early
 		// data.
 		// Signal early data if we're going to do it
-		if state.Config.AllowEarlyData {
+		if state.Config.AllowEarlyData && state.helloRetryRequest == nil {
 			state.Params.ClientSendingEarlyData = true
 			ed = &EarlyDataExtension{}
 			err = ch.Extensions.Add(ed)
@@ -423,7 +423,7 @@ func (state clientStateWaitSH) Next(hr handshakeMessageReader) (HandshakeState, 
 			cookie:            serverCookie.Cookie,
 			firstClientHello:  firstClientHello,
 			helloRetryRequest: hm,
-		}, nil, AlertNoAlert
+		}, []HandshakeAction{ResetOut{1}}, AlertNoAlert
 	}
 
 	// This is SH.
--- a/client-state-machine.go
+++ b/client-state-machine.go
@@ -58,7 +58,7 @@ type clientStateStart struct {
 	cookie            []byte
 	firstClientHello  *HandshakeMessage
 	helloRetryRequest *HandshakeMessage
-	hsCtx             HandshakeContext
+	hsCtx             *HandshakeContext
 }
 
 var _ HandshakeState = &clientStateStart{}
@@ -172,8 +172,10 @@ func (state clientStateStart) Next(hr handshakeMessageReader) (HandshakeState, [
 		}
 		ch.CipherSuites = compatibleSuites
 
+		// TODO(ekr@rtfm.com): Check that the ticket can be used for early
+		// data.
 		// Signal early data if we're going to do it
-		if len(state.Opts.EarlyData) > 0 {
+		if state.Config.AllowEarlyData {
 			state.Params.ClientSendingEarlyData = true
 			ed = &EarlyDataExtension{}
 			err = ch.Extensions.Add(ed)
@@ -255,9 +257,6 @@ func (state clientStateStart) Next(hr handshakeMessageReader) (HandshakeState, [
 		earlyTrafficSecret := deriveSecret(params, earlySecret, labelEarlyTrafficSecret, chHash)
 		logf(logTypeCrypto, "early traffic secret: [%d] %x", len(earlyTrafficSecret), earlyTrafficSecret)
 		clientEarlyTrafficKeys = makeTrafficKeys(params, earlyTrafficSecret)
-	} else if len(state.Opts.EarlyData) > 0 {
-		logf(logTypeHandshake, "[ClientStateWaitSH] Early data without PSK")
-		return nil, nil, AlertInternalError
 	} else {
 		clientHello, err = state.hsCtx.hOut.HandshakeMessageFromBody(ch)
 		if err != nil {
@@ -291,7 +290,6 @@ func (state clientStateStart) Next(hr handshakeMessageReader) (HandshakeState, [
 	if state.Params.ClientSendingEarlyData {
 		toSend = append(toSend, []HandshakeAction{
 			RekeyOut{epoch: EpochEarlyData, KeySet: clientEarlyTrafficKeys},
-			SendEarlyData{},
 		}...)
 	}
 
@@ -302,7 +300,7 @@ type clientStateWaitSH struct {
 	Config     *Config
 	Opts       ConnectionOptions
 	Params     ConnectionParameters
-	hsCtx      HandshakeContext
+	hsCtx      *HandshakeContext
 	OfferedDH  map[NamedGroup][]byte
 	OfferedPSK PreSharedKey
 	PSK        []byte
@@ -412,6 +410,11 @@ func (state clientStateWaitSH) Next(hr handshakeMessageReader) (HandshakeState, 
 			body:    h.Sum(nil),
 		}
 
+		state.hsCtx.receivedEndOfFlight()
+
+		// TODO(ekr@rtfm.com): Need to rekey with cleartext if we are on 0-RTT
+		// mode. In DTLS, we also need to bump the sequence number.
+		// This is a pre-existing defect in Mint. Issue #175.
 		logf(logTypeHandshake, "[ClientStateWaitSH] -> [ClientStateStart]")
 		return clientStateStart{
 			Config:            state.Config,
@@ -515,7 +518,6 @@ func (state clientStateWaitSH) Next(hr handshakeMessageReader) (HandshakeState, 
 	logf(logTypeCrypto, "master secret: [%d] %x", len(masterSecret), masterSecret)
 
 	serverHandshakeKeys := makeTrafficKeys(params, serverHandshakeTrafficSecret)
-
 	logf(logTypeHandshake, "[ClientStateWaitSH] -> [ClientStateWaitEE]")
 	nextState := clientStateWaitEE{
 		Config:                       state.Config,
@@ -530,13 +532,20 @@ func (state clientStateWaitSH) Next(hr handshakeMessageReader) (HandshakeState, 
 	toSend := []HandshakeAction{
 		RekeyIn{epoch: EpochHandshakeData, KeySet: serverHandshakeKeys},
 	}
+	// We're definitely not going to have to send anything with
+	// early data.
+	if !state.Params.ClientSendingEarlyData {
+		toSend = append(toSend, RekeyOut{epoch: EpochHandshakeData,
+			KeySet: makeTrafficKeys(params, clientHandshakeTrafficSecret)})
+	}
+
 	return nextState, toSend, AlertNoAlert
 }
 
 type clientStateWaitEE struct {
 	Config                       *Config
 	Params                       ConnectionParameters
-	hsCtx                        HandshakeContext
+	hsCtx                        *HandshakeContext
 	cryptoParams                 CipherSuiteParams
 	handshakeHash                hash.Hash
 	masterSecret                 []byte
@@ -596,6 +605,14 @@ func (state clientStateWaitEE) Next(hr handshakeMessageReader) (HandshakeState, 
 
 	state.handshakeHash.Write(hm.Marshal())
 
+	toSend := []HandshakeAction{}
+
+	if state.Params.ClientSendingEarlyData && !state.Params.UsingEarlyData {
+		// We didn't get 0-RTT, so rekey to handshake.
+		toSend = append(toSend, RekeyOut{epoch: EpochHandshakeData,
+			KeySet: makeTrafficKeys(state.cryptoParams, state.clientHandshakeTrafficSecret)})
+	}
+
 	if state.Params.UsingPSK {
 		logf(logTypeHandshake, "[ClientStateWaitEE] -> [ClientStateWaitFinished]")
 		nextState := clientStateWaitFinished{
@@ -608,7 +625,7 @@ func (state clientStateWaitEE) Next(hr handshakeMessageReader) (HandshakeState, 
 			clientHandshakeTrafficSecret: state.clientHandshakeTrafficSecret,
 			serverHandshakeTrafficSecret: state.serverHandshakeTrafficSecret,
 		}
-		return nextState, nil, AlertNoAlert
+		return nextState, toSend, AlertNoAlert
 	}
 
 	logf(logTypeHandshake, "[ClientStateWaitEE] -> [ClientStateWaitCertCR]")
@@ -622,13 +639,13 @@ func (state clientStateWaitEE) Next(hr handshakeMessageReader) (HandshakeState, 
 		clientHandshakeTrafficSecret: state.clientHandshakeTrafficSecret,
 		serverHandshakeTrafficSecret: state.serverHandshakeTrafficSecret,
 	}
-	return nextState, nil, AlertNoAlert
+	return nextState, toSend, AlertNoAlert
 }
 
 type clientStateWaitCertCR struct {
 	Config                       *Config
 	Params                       ConnectionParameters
-	hsCtx                        HandshakeContext
+	hsCtx                        *HandshakeContext
 	cryptoParams                 CipherSuiteParams
 	handshakeHash                hash.Hash
 	masterSecret                 []byte
@@ -706,7 +723,7 @@ func (state clientStateWaitCertCR) Next(hr handshakeMessageReader) (HandshakeSta
 type clientStateWaitCert struct {
 	Config        *Config
 	Params        ConnectionParameters
-	hsCtx         HandshakeContext
+	hsCtx         *HandshakeContext
 	cryptoParams  CipherSuiteParams
 	handshakeHash hash.Hash
 
@@ -760,7 +777,7 @@ func (state clientStateWaitCert) Next(hr handshakeMessageReader) (HandshakeState
 type clientStateWaitCV struct {
 	Config        *Config
 	Params        ConnectionParameters
-	hsCtx         HandshakeContext
+	hsCtx         *HandshakeContext
 	cryptoParams  CipherSuiteParams
 	handshakeHash hash.Hash
 
@@ -861,7 +878,7 @@ func (state clientStateWaitCV) Next(hr handshakeMessageReader) (HandshakeState, 
 
 type clientStateWaitFinished struct {
 	Params        ConnectionParameters
-	hsCtx         HandshakeContext
+	hsCtx         *HandshakeContext
 	cryptoParams  CipherSuiteParams
 	handshakeHash hash.Hash
 
@@ -933,6 +950,7 @@ func (state clientStateWaitFinished) Next(hr handshakeMessageReader) (HandshakeS
 	toSend := []HandshakeAction{}
 
 	if state.Params.UsingEarlyData {
+		logf(logTypeHandshake, "Sending end of early data")
 		// Note: We only send EOED if the server is actually going to use the early
 		// data.  Otherwise, it will never see it, and the transcripts will
 		// mismatch.
@@ -942,10 +960,11 @@ func (state clientStateWaitFinished) Next(hr handshakeMessageReader) (HandshakeS
 
 		state.handshakeHash.Write(eoedm.Marshal())
 		logf(logTypeCrypto, "input to handshake hash [%d]: %x", len(eoedm.Marshal()), eoedm.Marshal())
-	}
 
-	clientHandshakeKeys := makeTrafficKeys(state.cryptoParams, state.clientHandshakeTrafficSecret)
-	toSend = append(toSend, RekeyOut{epoch: EpochHandshakeData, KeySet: clientHandshakeKeys})
+		// And then rekey to handshake
+		toSend = append(toSend, RekeyOut{epoch: EpochHandshakeData,
+			KeySet: makeTrafficKeys(state.cryptoParams, state.clientHandshakeTrafficSecret)})
+	}
 
 	if state.Params.UsingClientAuth {
 		// Extract constraints from certicateRequest
@@ -1044,6 +1063,8 @@ func (state clientStateWaitFinished) Next(hr handshakeMessageReader) (HandshakeS
 		RekeyIn{epoch: EpochApplicationData, KeySet: serverTrafficKeys},
 		RekeyOut{epoch: EpochApplicationData, KeySet: clientTrafficKeys},
 	}...)
+
+	state.hsCtx.receivedEndOfFlight()
 
 	logf(logTypeHandshake, "[ClientStateWaitFinished] -> [StateConnected]")
 	nextState := stateConnected{
--- a/common_test.go
+++ b/common_test.go
@@ -17,7 +17,7 @@ func unhex(h string) []byte {
 	return b
 }
 
-func assert(t *testing.T, test bool, msg string) {
+func assertTrue(t *testing.T, test bool, msg string) {
 	t.Helper()
 	prefix := string("")
 	for i := 1; ; i++ {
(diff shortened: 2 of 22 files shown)


conn_test.go
```diff
@@ -274,55 +333,81 @@ func computeExporter(t *testing.T, c *Conn, label string, context []byte, length
...
+	assertByteEquals(t, computeExporter(t, client, "E", emptyContext, 20), computeExporter(t, server, "E", emptyContext, 20))
+	assertNotByteEquals(t, computeExporter(t, client, "E", emptyContext, 20), computeExporter(t, server, "E", emptyContext, 21))
+	assertNotByteEquals(t, computeExporter(t, client, "E", emptyContext, 20), computeExporter(t, server, "F", emptyContext, 20))
+	assertByteEquals(t, computeExporter(t, client, "E", []byte{'A'}, 20), computeExporter(t, server, "E", []byte{'A'}, 20))
+	assertNotByteEquals(t, computeExporter(t, client, "E", []byte{'A'}, 20), computeExporter(t, server, "E", []byte{'B'}, 20))
+}
+
+func testConnInner(t *testing.T, name string, p testInstanceState) {
+	// Configs array:
+	configs := map[string]*Config{"basic config": basicConfig,
```
bifurcation: Nit: New line before "basic config"

common_test.go
```diff
@@ -81,12 +81,81 @@ func assertCipherSuiteParamsEquals(t *testing.T, a, b CipherSuiteParams) {
...
+				f(t, n, state)
+			})
+			continue
+		}
+		runParametrizedInner(t, n, state, next, f)
+	}
+}
+
+// Nominally public API.
+type testParameter struct {
```
bifurcation: It seems like instead of having this struct and a bunch of `[]testParameter`, you could just do something like:

`type testParams map[string][]string`

common_test.go
```diff
@@ -81,12 +81,81 @@ func assertCipherSuiteParamsEquals(t *testing.T, a, b CipherSuiteParams) {
...
+// Utilities for parametrized tests
+// Represents the configuration for a given test instance.
+type testInstanceState interface {
+	set(string, string)
+	get(string) string
+}
+
+// This complicated mixin stuff is in case we want to extend this to
+// pass other parameters that aren't srings. Arguably YAGNI.
+type testInstanceStateMixin struct {
```
bifurcation: I would just leave this as a `map[string]string` for now. (`type testInstanceState map[string]string`)  If you want to let the test pass other stuff in, you could add an `interface{}` argument to the "public" API that the caller can use to pass context to each test.

common_test.go
```diff
@@ -81,12 +81,81 @@ func assertCipherSuiteParamsEquals(t *testing.T, a, b CipherSuiteParams) {
...
+// Nominally public API.
+type testParameter struct {
+	name string
+	vals []string
+}
+
+type parametrizedTest func(t *testing.T, name string, p testInstanceState)
+
+// This is the function you call.
+func runParametrizedTest(t *testing.T, inparams []testParameter, f parametrizedTest) {
```
bifurcation: I think I might need a little more explanation here.  Let me check my understanding here:

* `inparams` is basically a `map[string][]string`, where the key is the name of a parameter, and the array are its possible values
* For each parameter, you iterate through its values, running the test on a params struct with that value
* At the end of that loop, you leave that parameter on its last value, and move on to the next param

So given the following params dict:

```
{
  {"a", {"x", "y", "z"}},
  {"b", {"r", "s"}},
  {"c", {"q"}},
}
```

... you would go through the following cases

* a="x" b="" c=""
* a="y" b="" c=""
* a="z" b="" c=""
* a="z" b="" c=""
* a="z" b="r" c=""
* a="z" b="s" c=""
* a="z" b="s" c="q"

Assuming I've got all that right... That seems kind of odd, for two reasons.  First, you have the default values for things in the beginning, then you've got things stuck on their last value.

ISTM there are basically two sensible options here:
* Test all the combinations, or if you want to keep things linear
* Set a default value for each parameter, which is used outside of its test run

So with the above, you would have:

* a="x" b=default c=default
* a="y" b=default c=default
* a="z" b=default c=default
* a="z" b=default c=default
* a=default b="r" c=default
* a=default b="s" c=default
* a=default b=default c="q"

You could also treat "" as the default and leave it up to the test to deal with it.  In that case, the only real change you would need to do would be `state.set(param.name, "")` in `runParametrizedInner()`.

In any case, it would be good to have a brief explanation in a comment here about how this all works.

conn_test.go
```diff
@@ -86,14 +86,38 @@ func (p *pipeConn) RemoteAddr() net.Addr               { return nil }
 func (p *pipeConn) SetDeadline(t time.Time) error      { return nil }
 func (p *pipeConn) SetReadDeadline(t time.Time) error  { return nil }
 func (p *pipeConn) SetWriteDeadline(t time.Time) error { return nil }
+func (p *pipeConn) Left() int                          { return p.r.Len() }
```
bifurcation: Nit: Suggest renaming to `Len` to avoid confusion with Left/Right, and align with `bytes.Buffer`.

bifurcation: Or better yet, since you only ever test `Left() > 0`, just have a method `Empty() bool`

conn_test.go
```diff
@@ -123,28 +154,38 @@ func (b *bufferedConn) Flush() error {
...
+}
+
+func (b *bufferedConn) Clear() {
+	b.buffer.Reset()
+}
+
 func newBufferedConn(p net.Conn) *bufferedConn {
-	return &bufferedConn{bytes.Buffer{}, p}
+	return &bufferedConn{
+		false, bytes.Buffer{}, p, 0, make(map[int]bool, 0),
```
bifurcation: Please expand with parameter names.

conn_test.go
```diff
@@ -86,14 +86,38 @@ func (p *pipeConn) RemoteAddr() net.Addr               { return nil }
...
 func (p *pipeConn) SetWriteDeadline(t time.Time) error { return nil }
+func (p *pipeConn) Left() int                          { return p.r.Len() }
 
 type bufferedConn struct {
-	buffer bytes.Buffer
-	w      net.Conn
+	autoflush   bool
+	buffer      bytes.Buffer
+	w           net.Conn
+	ctr         int
```
bifurcation: `writeCounter` for readability

conn_test.go
```diff
@@ -86,14 +86,38 @@ func (p *pipeConn) RemoteAddr() net.Addr               { return nil }
...
+	lossPattern map[int]bool
```
bifurcation: `lostWrite`

conn_test.go
```diff
@@ -86,14 +86,38 @@ func (p *pipeConn) RemoteAddr() net.Addr               { return nil }
...
 }
 
-func (b *bufferedConn) Write(buf []byte) (n int, err error) {
-	return b.buffer.Write(buf)
+func (b *bufferedConn) Write(buf []byte) (int, error) {
+	ctr := b.ctr
+	b.ctr++
+	if ok := b.lossPattern[ctr]; ok {
```
bifurcation: You can just do `b.lossPattern[ctr]` here.  `bool` defaults to `false`.

conn_test.go
```diff
@@ -108,6 +132,13 @@ func (p *bufferedConn) RemoteAddr() net.Addr               { return nil }
 func (p *bufferedConn) SetDeadline(t time.Time) error      { return nil }
 func (p *bufferedConn) SetReadDeadline(t time.Time) error  { return nil }
 func (p *bufferedConn) SetWriteDeadline(t time.Time) error { return nil }
+func (b *bufferedConn) SetAutoflush() {
+	b.autoflush = true
+}
+func (b *bufferedConn) Left() int {
```
bifurcation: Same comment as above.

conn_test.go
```diff
@@ -123,28 +154,38 @@ func (b *bufferedConn) Flush() error {
...
 	serverKey, clientKey             crypto.Signer
 	serverCert, clientCert           *x509.Certificate
 	certificates, clientCertificates []*Certificate
+	clientName, serverName           string
 
 	psk  PreSharedKey
 	psks *PSKMapCache
 
-	basicConfig, dtlsConfig, nbConfig, hrrConfig, alpnConfig, pskConfig, pskECDHEConfig, pskDHEConfig, resumptionConfig, ffdhConfig, x25519Config *Config
+	basicConfig, dtlsConfig, nbConfig, nbConfigDTLS, hrrConfig, alpnConfig, pskConfig, pskDTLSConfig, pskECDHEConfig, pskDHEConfig, resumptionConfig, ffdhConfig, x25519Config *Config
```
bifurcation: `nbDTLSConfig`

conn_test.go
bifurcation: Running `gofmt -s` should fix any other similar stuff.

conn_test.go
```diff
@@ -775,42 +849,54 @@ func TestResumption(t *testing.T) {
...
+	sbConn := newBufferedConn(sConn)
+	sbConn.SetAutoflush()
+
+	client := Client(cbConn, &conf)
+	server := Server(sbConn, &conf)
+
+	client.Handshake() // This sends CH
+	zdata := []byte("ABC")
+	n, err := client.Write(zdata) // This should succeeed
+	assertNotError(t, err, "Client was able to write")
```
bifurcation: "Client was not able to write"

common_test.go
ekr: You could indeed, but when I have this many []s I tend to get lost, so I'd prefer to leave as-is unless you object.

client-state-machine.go
```diff
@@ -58,7 +58,7 @@ type clientStateStart struct {
 	cookie            []byte
 	firstClientHello  *HandshakeMessage
 	helloRetryRequest *HandshakeMessage
-	hsCtx             HandshakeContext
+	hsCtx             *HandshakeContext
```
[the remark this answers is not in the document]
ekr: The conn and the state share the hsCtx.

common_test.go
ekr: Ah, no. This is supposed to run all the combinations for each parameter. So, given the following test:

```
func testParametrized(t *testing.T, n string, s testInstanceState) {
	fmt.Println(n, s)
}

func TestParametrized(t *testing.T) {
	p := []testParameter{
		testParameter{"a", []string{"a1", "a2"}},
		testParameter{"b", []string{"b1", "b2"}},
		testParameter{"c", []string{"c1", "c2"}},
	}

	runParametrizedTest(t, p, testParametrized)
}

````

You get the output:

```
a=a1/b=b1/c=c1 &{{map[c:c1 a:a1 b:b1]}}
a=a1/b=b1/c=c2 &{{map[c:c2 a:a1 b:b1]}}
a=a1/b=b2/c=c1 &{{map[a:a1 b:b2 c:c1]}}
a=a1/b=b2/c=c2 &{{map[a:a1 b:b2 c:c2]}}
a=a2/b=b1/c=c1 &{{map[b:b1 c:c1 a:a2]}}
a=a2/b=b1/c=c2 &{{map[c:c2 a:a2 b:b1]}}
a=a2/b=b2/c=c1 &{{map[a:a2 b:b2 c:c1]}}
a=a2/b=b2/c=c2 &{{map[a:a2 b:b2 c:c2]}}
```

Perhaps you're getting confused by len(next), but that's about the length of the remaining parameter list to iterate over.

File: record-layer.go [hunk 14]
```diff
@@ -116,47 +132,55 @@ func (r *RecordLayer) Rekey(epoch Epoch, factory aeadFactory, key []byte, iv []b
...
-	var i int
-	for i = len(c.seq) - 1; i >= 0; i-- {
-		c.seq[i]++
-		if c.seq[i] != 0 {
-			break
-		}
-	}
-
-	if i < 0 {
+	if c.seq >= 1<<48 {
```
bifurcation: If you've got a 48-bit field, you should panic before you overflow, i.e., `c.seq >= (1 << 48) - 1`.

record-layer_test.go
```diff
@@ -25,41 +25,39 @@ func TestRekey(t *testing.T) {
...
 
 	key := unhex(keyHex)
 	iv := unhex(ivHex)
 
 	cs, err := newCipherStateAead(EpochApplicationData, newAESGCM, key, iv)
 	assertNotError(t, err, "Couldn't create cipher state")
-	for i := 0; i < sequenceNumberLen; i++ {
-		cs.seq[cs.ivLength-i-1] = 0xFF
-	}
+	cs.seq = 1 << 48
```
bifurcation: This is one off from the test before; cf. the comment about overflow above.

record-layer.go
```diff
@@ -318,28 +356,47 @@ func (r *RecordLayer) nextRecord() (*TLSPlaintext, error) {
...
 		// TODO(ekr@rtfm.com): Handle duplicates.
-		logf(logTypeIO, "RecordLayer.ReadRecord epoch=[%s] seq=[%x] [%d] ciphertext=[%x]", cipher.epoch.label(), seq, pt.contentType, pt.fragment)
+		seq, _ = decodeUint(header[3:11], 8)
+		epoch := Epoch(seq >> 48)
+
+		// Look up the cipher suite from the epoch
+		if epoch != cipher.epoch {
+			logf(logTypeIO, "%s Message from non-current epoch: [%v != %v]", r.label, epoch,
+				cipher.epoch)
+			if !allowOldEpoch {
```
bifurcation: It seems like you don't know at this point whether the epoch is old or just bogus.  Maybe reorder the invalid vs. old check and use the knowledge of the epoch as a sign that it's old, not bogus.

common_test.go
bifurcation: Yep, this is my error.  Thanks for the explanation.

record-layer.go
```diff
@@ -116,47 +132,55 @@ func (r *RecordLayer) Rekey(epoch Epoch, factory aeadFactory, key []byte, iv []b
...
 	r.cipher = cipher
+	if r.datagram && r.direction == directionRead {
+		r.readCiphers[epoch] = cipher
+	}
 	return nil
 }
 
-func (c *cipherState) formatSeq(datagram bool) []byte {
-	seq := append([]byte{}, c.seq...)
+func (r *RecordLayer) DiscardReadKey(epoch Epoch) {
```
bifurcation: This function doesn't appear to be called anywhere.

conn.go
```diff
@@ -570,62 +609,6 @@ func (c *Conn) takeAction(actionGeneric HandshakeAction) Alert {
...
 
-	case SendEarlyData:
-		logf(logTypeHandshake, "%s Sending early data...", label)
-		_, err := c.Write(c.EarlyData)
-		if err != nil {
-			logf(logTypeHandshake, "%s Error writing early data: %v", label, err)
-			return AlertInternalError
-		}
-
-	case ReadPastEarlyData:
```
bifurcation: I was slightly happier having this work done outside the HandshakeState, which preserves the property that the handshake states' `Next()` methods only have to deal with handshake messages.  It also seems like it saves you the extra state value `ReadPastEarlyData`; you just have the action list returned from `Negotiated.Next()` indicate whether you should read past early data or not.  What's the rationale for moving this into the `Next()` methods?

record-layer_test.go
```diff
@@ -131,15 +129,15 @@ func TestDecryptRecord(t *testing.T) {
 	ciphertext2 := unhex(ciphertext2Hex)
 
 	// Test successful decrypt
-	r := NewRecordLayerTLS(bytes.NewBuffer(ciphertext1))
+	r := NewRecordLayerTLS(bytes.NewBuffer(ciphertext1), directionRead)
```
bifurcation: This file might benefit from a `newRecordLayerFromBytes`, as in `handshake-layer_test.go`

dtls.go
```diff
@@ -26,3 +47,180 @@ func dtlsConvertVersion(version uint16) uint16 {
...
+	if !h.hIn.datagram {
+		return
+	}
+
+	// Disarm the ACK timer
+	h.timers.cancel(ackTimerLabel)
+
+	// But send an ACK immediately.
+	h.sendAck()
+
```
bifurcation: Nit: Extra blank line

conn_test.go
```diff
@@ -775,42 +849,54 @@ func TestResumption(t *testing.T) {
...
-	assertByteEquals(t, client.state.serverTrafficSecret, server.state.serverTrafficSecret)
-	assert(t, client.state.Params.UsingEarlyData, "Session did not negotiate early data")
-	assertByteEquals(t, client.EarlyData, server.EarlyData)
+func Test0xRTT(t *testing.T) {
+	params := []testParameter{
+		testParameter{
+			"dtls",
+			[]string{"true", "false"},
+		},
+	}
```
bifurcation: If you change from `[]testParameter` to `type testParams map[string][]string`, then you can write things like this much more compactly:

```
testParams{
  "dtls": {"true", "false"},
}
```

conn_test.go
```diff
@@ -1190,3 +1275,373 @@ func TestDTLS(t *testing.T) {
...
+
+	// Only client should be running a timer.
+	waiting, timeout := server.GetDTLSTimeout()
+	assertTrue(t, !waiting, fmt.Sprintf("Server timer armed: %v", timeout))
+
+	waiting, timeout = client.GetDTLSTimeout()
+	assertTrue(t, waiting, "Client timer not armed")
+
+	// Now check the timer.
+	time.Sleep(timeout)
```
bifurcation: You should be careful with this; IIRC go times out tests after a few seconds.

It might be worth considering moving away from using `time.Now()` directly in the timer code.  That would enable you to just advance time forward as you like, so that you don't have to sleep.  Boulder uses https://github.com/jmhodges/clock for this purpose. 

However, synthetic clocks (including the one above) typically don't try to implement a parallel to `time.Timer`, so maybe there's a choice here between using `time.Timer + time.Now() +time.Sleep()`, or rolling our own timer class and using a synthetic clock.

conn_test.go
```diff
@@ -1190,3 +1275,373 @@ func TestDTLS(t *testing.T) {
...
+	assertEquals(t, server.GetHsState(), StateServerConnected)
+
+	assertDeepEquals(t, client.state.Params, server.state.Params)
+	assertCipherSuiteParamsEquals(t, client.state.cryptoParams, server.state.cryptoParams)
+	assertByteEquals(t, client.state.resumptionSecret, server.state.resumptionSecret)
+	assertByteEquals(t, client.state.clientTrafficSecret, server.state.clientTrafficSecret)
+	assertByteEquals(t, client.state.serverTrafficSecret, server.state.serverTrafficSecret)
+}
+
+func checkTimers(t *testing.T, c *Conn, labels []string) {
```
bifurcation: A little more verbosity in the names here might be nice.
```
checkTimers -> checkTimersEqualLabels
armed -> timers
ma -> timerForLabel
mb -> labelForTimer
a -> timer || label
```

conn.go
```diff
@@ -788,6 +781,31 @@ func (c *Conn) Handshake() Alert {
 		if connected {
 			c.state = state.(stateConnected)
 			c.handshakeComplete = true
+
+			if !c.isClient {
+				// Send NewSessionTicket if configured to
+				if c.config.SendSessionTickets {
```
bifurcation: Combine this `if` statement with the previous.
```
if !c.isClient && c.config.SendSessionTickets {...}
```

negotiation.go
```diff
@@ -168,10 +168,10 @@ func CertificateSelection(serverName *string, signatureSchemes []SignatureScheme
 	return nil, 0, fmt.Errorf("No certificates compatible with signature schemes")
 }
 
-func EarlyDataNegotiation(usingPSK, gotEarlyData, allowEarlyData bool) bool {
+func EarlyDataNegotiation(usingPSK, gotEarlyData, allowEarlyData bool) ( /* using */ bool /* rejected */, bool) {
```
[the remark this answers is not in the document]
ekr: OK. I kind of find that side effect in Go odd, but OK.

timer.go
```diff
@@ -0,0 +1,122 @@
...
+
+import (
+	"time"
+)
+
+// This is a simple timer implementation. Timers are stored in a sorted
+// list.
+type timerCb func() error
+
+type timer struct {
```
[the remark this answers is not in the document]
ekr: Yes, these timers are uncoupled from the system clock.

state-machine_test.go
```diff
@@ -173,6 +173,8 @@ func TestStateMachineIntegration(t *testing.T) {
 				},
 			},
 
+			/* Commented out because PeekRecordType() not available without a record layer
```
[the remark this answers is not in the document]
ekr: Sure.

server-state-machine.go
```diff
@@ -813,10 +851,44 @@ func (state serverStateWaitEOED) Next(hr handshakeMessageReader) (HandshakeState
 	return waitFlight2, toSend, AlertNoAlert
 }
 
+var _ HandshakeState = &serverStateReadPastEarlyData{}
+
+type serverStateReadPastEarlyData struct {
```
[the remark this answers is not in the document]
ekr: Right. That's what this code is supposed to do. Do you think it's wrong or you just want better comments?

conn_test.go
ekr: Right. I wrote my own timer class so I could do a synthetic clock and then got lazy.

conn.go
ekr: This isn't right, because the early data copy happens inside !c.isClient.

ekr:
You asked about moving the new states into .Next(). My claim is that they are part of the state machine, so I moved them, and it also has to be able to re-enter after WouldBlock and we already have machinery for that.

server-state-machine.go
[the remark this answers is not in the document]
bifurcation: The latter.  The above was just me working out what I thought the change to the state diagram comment should be.

conn_test.go
bifurcation: In that case, maybe add `TODO(ekr@rtfm.com): Make these independent of the system clock`.

conn.go
bifurcation: If that's your intent, then you need to move the early data copy up into `!c.isClient`.  Right now, it's parallel.

bifurcation:
OK, reentering after WouldBlock seems like a good enough reason.

ekr:
Do you think you could help me with the state diagram?

ekr:
WRT nextState for WaitEOED. The issue is you also need to queue up the key changes, so I'm not sure how much it helps.

ekr:
@bifurcation: PTAL. This might need one more once-over before landing, but I think we're close

dtls.go
```diff
@@ -26,3 +47,180 @@ func dtlsConvertVersion(version uint16) uint16 {
...
+	// Reset the timer
+	h.timers.start(retransmitTimerLabel,
+		h.handshakeRetransmit,
+		h.timeoutMS)
+
+	return nil
+}
+
+type connTimerCb func(c *Conn) error
```
bifurcation: This type does not appear to be used

dtls.go
```diff
@@ -2,14 +2,35 @@ package mint
 
 import (
 	"fmt"
+	"github.com/bifurcation/mint/syntax"
+	"time"
 )
 
 // This file is a placeholder. DTLS-specific stuff (timer management,
 // ACKs, retransmits, etc. will eventually go here.
```
bifurcation: We can probably delete this comment now :)

timer.go
```diff
@@ -0,0 +1,122 @@
+package mint
+
+import (
+	"time"
+)
+
+// This is a simple timer implementation. Timers are stored in a sorted
+// list.
+// TODO(ekr@rtfm.com): Add a way to uncouple these from the system
+// clock.
```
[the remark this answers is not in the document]
ekr: This isn't trivial, and this is a huge CL, so let's leave it for later.

server-state-machine.go
```diff
@@ -239,10 +239,6 @@ func (state serverStateStart) Next(hr handshakeMessageReader) (HandshakeState, [
 			logf(logTypeHandshake, "[ServerStateStart] Error in PSK negotiation [%v]", err)
 			return nil, nil, AlertInternalError
 		}
-		if clientSentCookie && initialCipherSuite.Suite != params.Suite {
```
bifurcation: Why do you think this check is not necessary?

>  Servers MUST ensure that they negotiate the same cipher suite when receiving a conformant updated ClientHello

Are you assuming that this happens automatically?

state-machine.go
```diff
@@ -27,6 +27,10 @@ type RekeyOut struct {
 	KeySet keySet
 }
 
+type ResetOut struct {
+	seq uint64
```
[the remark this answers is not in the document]
ekr: I was trying to be general, but I think it's actually always 1

server-state-machine.go
ekr: I believe it is done here: https://github.com/bifurcation/mint/blob/master/server-state-machine.go#L253

state-machine.go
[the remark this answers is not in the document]
bifurcation: Ok, we can leave it for now.

PR #191, opened 2018-03-17
Most minimal possible draft-23 update

ekr:


marten-seemann:
I'm not following the TLS draft changes closely, but according to the changelog, there were more changes than these for draft-23. Are all of these optional?

ekr:
Sort of. They're supposed to help you guide certificate selection, so it's not formally correct, but as a practical matter, this will get interop. Happy to file an issue for it.

PR #192, opened 2018-04-11
Require client provide certs if RequireClientAuth=true

jared2501:


jared2501:
Hey @bifurcation, before I look into the tests for this, does this look right? The issue I'm trying to fix is when a server requires client certificates, but the client doesn't send any, we let the TLS session establish when we shouldn't.

marten-seemann:
The spec doesn’t require the client to send a certificate if the server requests one. I think the right way to deal with this is as the standard library does.

jared2501:
@marten-seemann - ah hmm, perhaps RequireClientAuth is a bit of a misnomre then? Also, when you say "as the standard library does", do you mean by allowing the client to set different levels from NoClientCert to RequireAndVerifyClientCert?

PR #193, opened 2018-04-11
WIP: SPAKE2

bifurcation:
This patch adds a PAKE key exchange mechanism using more or less the same mechanism proposed in [draft-barnes-tls-pake](https://github.com/bifurcation/tls-pake/blob/master/draft-barnes-tls-pake.md).  The only major difference is that the extension syntax is slightly modified to allow the client to offer multiple identities.


bifurcation:
The document has now been updated to match this implementation.

PR #195, opened 2018-05-15
Fix key share pub/priv order in test.

### LINKED ISSUE #144: Key Share tests failing

martinthomson:
Fixes #144.

### DIFF
--- a/crypto_test.go
+++ b/crypto_test.go
@@ -65,7 +65,7 @@ func TestNewKeyShare(t *testing.T) {
 	}
 
 	for _, group := range nonECGroups {
-		priv, pub, err := newKeyShare(group)
+		pub, priv, err := newKeyShare(group)
 		assertNotError(t, err, "Failed to generate new key pair")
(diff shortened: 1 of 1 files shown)


martinthomson:
That CI change looks like a pre-existing problem in the DTLS tests.  I blame @ekr.

martinthomson:
Oh, forgot to add.  `go test -run TestNewKeyShare -count 1000` fails with the old code, but completes - slowly - with this change.

PR #196, opened 2018-05-16
TLS 1.3 Final

martinthomson:
This should make mint interoperable for draft-28 changes in *TLS*.  I can take a swing at implementing the record layer changes for DTLS, but that is a far bigger change and I'd like to do that separately.  (Either that, or get @ekr to do those because I made the changes in NSS.)

### DIFF
--- a/common.go
+++ b/common.go
@@ -6,7 +6,7 @@ import (
 )
 
 const (
-	supportedVersion  uint16 = 0x0304
+	tls13Version      uint16 = 0x0304
 	tls12Version      uint16 = 0x0303
 	tls10Version      uint16 = 0x0301
 	dtls12WireVersion uint16 = 0xfefd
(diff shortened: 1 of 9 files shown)


quite:
I tried this building this and succeeded running `mint-client-https -url https://tls13.crypto.mozilla.org:443`

marten-seemann:
We should merge this soon.
@martinthomson: Any idea why the tests are failing on the CI? They're passing locally, but we should probably fix that before merging.

martinthomson:
No idea why this is failing, it's DTLS (and the DTLS code is not right for other reasons).

common.go
```diff
@@ -6,7 +6,7 @@ import (
 )
 
 const (
-	supportedVersion  uint16 = 0x7f16 // draft-22
+	supportedVersion  uint16 = 0x0304
```
bifurcation: Let's go ahead and change this to `tls13version`

PR #198, opened 2018-09-02
Enable fields to be marked for omission from a struct

bifurcation:
This way you can have structs that marshal / unmarshal as TLS, but also have extra information that is produced locally.  The proposed syntax is:

```
type Foo struct {
  A uint16
  B uint16 `tls:"omit"`
  C uint16
}
```

### DIFF
--- a/syntax/decode.go
+++ b/syntax/decode.go
@@ -102,6 +102,13 @@ func newTypeDecoder(t reflect.Type) decoderFunc {
 
 ///// Specific decoders below
 
+func omitDecoder(d *decodeState, v reflect.Value, opts decOpts) int {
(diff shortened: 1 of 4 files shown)


ekr:
Why not just use lower-case for this? Those fields are inaccessible to reflection anyway.

For instance: https://github.com/ekr/minq/blob/master/codec.go#L155

bifurcation:
Renewing this PR.  @ekr - you could use private fields for this, but I don't see a need to conflate the two ideas. `encoding/json`, for example, makes exactly this distinction, allowing a `"-"` annotation for a public field you want to omit in the JSON encoding.

PR #199, opened 2018-11-03
QUIC record layer changes

martinthomson:
This rolls up all the changes for draft-28 as well.

Needs review though.

### DIFF
--- a/crypto.go
+++ b/crypto.go
@@ -30,11 +30,10 @@ var prng = rand.Reader
 type AeadFactory func(key []byte) (cipher.AEAD, error)
 
 type CipherSuiteParams struct {
-	Suite  CipherSuite
-	Cipher AeadFactory // Cipher factory
-	Hash   crypto.Hash // Hash function
-	KeyLen int         // Key length in octets
-	IvLen  int         // IV length in octets
+	Suite      CipherSuite
+	Cipher     AeadFactory    // Cipher factory
+	Hash       crypto.Hash    // Hash function
+	KeyLengths map[string]int // This maps keys (the label used for HKDF-Expand-Label) to the length of the key needed.
 }
 
 type signatureAlgorithm uint8
@@ -91,18 +90,16 @@ var (
 
 	cipherSuiteMap = map[CipherSuite]CipherSuiteParams{
 		TLS_AES_128_GCM_SHA256: {
-			Suite:  TLS_AES_128_GCM_SHA256,
-			Cipher: newAESGCM,
-			Hash:   crypto.SHA256,
-			KeyLen: 16,
-			IvLen:  12,
+			Suite:      TLS_AES_128_GCM_SHA256,
+			Cipher:     newAESGCM,
+			Hash:       crypto.SHA256,
+			KeyLengths: map[string]int{labelForKey: 16, labelForIV: 12},
 		},
 		TLS_AES_256_GCM_SHA384: {
-			Suite:  TLS_AES_256_GCM_SHA384,
-			Cipher: newAESGCM,
-			Hash:   crypto.SHA384,
-			KeyLen: 32,
-			IvLen:  12,
+			Suite:      TLS_AES_256_GCM_SHA384,
+			Cipher:     newAESGCM,
+			Hash:       crypto.SHA384,
+			KeyLengths: map[string]int{labelForKey: 32, labelForIV: 12},
 		},
 	}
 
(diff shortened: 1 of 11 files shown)


crypto.go
```diff
@@ -604,18 +606,20 @@ func computeFinishedData(params CipherSuiteParams, baseKey []byte, input []byte)
...
 	logf(logTypeCrypto, "making traffic keys: secret=%x", secret)
-	return keySet{
-		cipher: params.Cipher,
-		key:    HkdfExpandLabel(params.Hash, secret, "key", []byte{}, params.KeyLen),
-		iv:     HkdfExpandLabel(params.Hash, secret, "iv", []byte{}, params.IvLen),
+	return KeySet{
+		Cipher: params.Cipher,
+		Key:    HkdfExpandLabel(params.Hash, secret, "key", []byte{}, params.KeyLen),
+		Iv:     HkdfExpandLabel(params.Hash, secret, "iv", []byte{}, params.IvLen),
+		Pn:     HkdfExpandLabel(params.Hash, secret, "pn", []byte{}, params.KeyLen),
```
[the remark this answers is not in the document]
ekr: I agree with this point, but in the interest of getting done, I think we ought to just land this and file an issue.

crypto.go
```diff
@@ -27,11 +27,11 @@ import (
 
 var prng = rand.Reader
 
-type aeadFactory func(key []byte) (cipher.AEAD, error)
+type AeadFactory func(key []byte) (cipher.AEAD, error)
```
bifurcation: `AEADFactory` please

crypto.go
```diff
@@ -604,18 +606,20 @@ func computeFinishedData(params CipherSuiteParams, baseKey []byte, input []byte)
...
 }
 
-type keySet struct {
-	cipher aeadFactory
-	key    []byte
-	iv     []byte
+type KeySet struct {
+	Cipher AeadFactory
+	Key    []byte
+	Iv     []byte
```
bifurcation: `IV`

crypto.go
```diff
@@ -546,8 +546,10 @@ const (
 //    opaque label<9..255>;
 //    opaque hash_value<0..255>;
 // };
+var HkdfLabelPrefix = "tls13 "
```
bifurcation: `HKDFLabelPrefix`

bifurcation: Is the idea that if you were doing QUIC or something, you would do something like `mint.HKDFLabelPrefix = "quic "`?  Does that actually work?

crypto.go
```diff
@@ -604,18 +606,20 @@ func computeFinishedData(params CipherSuiteParams, baseKey []byte, input []byte)
...
 
-type keySet struct {
-	cipher aeadFactory
-	key    []byte
-	iv     []byte
+type KeySet struct {
+	Cipher AeadFactory
+	Key    []byte
+	Iv     []byte
+	Pn     []byte
```
bifurcation: What is this thing?  If it's some QUIC-specific thing, I'm OK having it here if it's necessary, but please include a comment explaining it.

handshake-layer.go
```diff
@@ -394,9 +401,13 @@ func (h *HandshakeLayer) ReadMessage() (*HandshakeMessage, error) {
 }
 
 func (h *HandshakeLayer) QueueMessage(hm *HandshakeMessage) error {
-	hm.cipher = h.conn.cipher
-	h.queued = append(h.queued, hm)
-	return nil
+	if h.datagram {
+		hm.cipher = h.conn.(*RecordLayerImpl).cipher
```
bifurcation: What's the point of the indirection of the record layer if the calling code is always assuming that it has the default impl?

record-layer.go
```diff
@@ -50,10 +62,28 @@ type cipherState struct {
...
+	Rekey(epoch Epoch, factory AeadFactory, keys *KeySet) error
+	ResetClear(seq uint64)
+	DiscardReadKey(epoch Epoch)
+	PeekRecordType(block bool) (RecordType, error)
+	ReadRecord() (*TLSPlaintext, error)
+	WriteRecord(pt *TLSPlaintext) error
+	Epoch() Epoch
+}
+
+type RecordLayerImpl struct {
```
bifurcation: I would prefer something more descriptive, e.g., `DefaultRecordLayer`.  Since the whole premise here is that there can be other impls.

bifurcation: We should also actually make this the default impl, so that if a record layer isn't specified, this one gets filled in.  Unclear to me from this PR whether this is already the case.

crypto.go
martinthomson: That wouldn't work because you only want to use the "quic " label in certain limited cases.  The rest of the key schedule keeps the "tls13 " label.  It's stupendously inconsistent.  For which I blame ekr.

crypto.go
martinthomson: I think that this object should move to `map[[]byte][]byte` or `map[string][]byte` and have the caller pass in a list of the keys that it wants generated.

martinthomson: I just implemented this change.  It's pretty nice.

record-layer.go
martinthomson: Yeah, this looks like it is wrong to me.  The number of casts down to `DefaultRecordLayer` are too high.  If the intent is to abstract this out, then this set of changes falls short.  I've manage to fix the other comments, but this one is tricky.

martinthomson:
Hmm, it's been a while, but I think that I addressed the comments, aside from the one.

PR #206, opened 2019-09-01
Enable struct fields to be skipped in encode/decode

bifurcation:
Add a `none` annotation to allow the encoder / decoder to skip struct fields.  Works the same as the comparable annotation in `encoding/json`.


bifurcation:
Closing as a dup of #198

PR #208, opened 2019-09-01
Support raw public keys (RFC 7520)

bifurcation:
Main components here:

* Definition of the `client_certificate_type` and `server_certificate_type` extensions
* Related negotiation logic
* Generalization of certificate logic in the handshake to support both raw keys and certificates

### DIFF
--- a/client-state-machine.go
+++ b/client-state-machine.go
@@ -102,6 +102,24 @@ func (state clientStateStart) Next(hr handshakeMessageReader) (HandshakeState, [
 		alpn = &ALPNExtension{Protocols: state.Opts.NextProtos}
 	}
 
+	// Raw public keys
+	var cct *ClientCertTypeExtension
+	var sct *ClientCertTypeExtension
+	if state.Config.AllowRawPublicKeys || !state.Config.ForbidCertificates {
+		cct := &ClientCertTypeExtension{HandshakeType: HandshakeTypeClientHello}
+		sct := &ServerCertTypeExtension{HandshakeType: HandshakeTypeClientHello}
+
+		if state.Config.AllowRawPublicKeys {
+			cct.CertificateTypes = append(cct.CertificateTypes, CertificateTypeRawPublicKey)
+			sct.CertificateTypes = append(sct.CertificateTypes, CertificateTypeRawPublicKey)
+		}
+
+		if !state.Config.ForbidCertificates {
+			cct.CertificateTypes = append(cct.CertificateTypes, CertificateTypeX509)
+			sct.CertificateTypes = append(sct.CertificateTypes, CertificateTypeX509)
+		}
+	}
+
 	// Construct base ClientHello
 	ch := &ClientHelloBody{
 		LegacyVersion: wireVersion(state.hsCtx.hIn),
@@ -113,7 +131,7 @@ func (state clientStateStart) Next(hr handshakeMessageReader) (HandshakeState, [
 		return nil, nil, AlertInternalError
 	}
 	for _, ext := range []ExtensionBody{&sv, &sni, &ks, &sg, &sa} {
-		err := ch.Extensions.Add(ext)
+		err = ch.Extensions.Add(ext)
 		if err != nil {
 			logf(logTypeHandshake, "[ClientStateStart] Error adding extension type=[%v] [%v]", ext.Type(), err)
 			return nil, nil, AlertInternalError
@@ -125,6 +143,18 @@ func (state clientStateStart) Next(hr handshakeMessageReader) (HandshakeState, [
 		err := ch.Extensions.Add(alpn)
 		if err != nil {
 			logf(logTypeHandshake, "[ClientStateStart] Error adding ALPN extension [%v]", err)
+			return nil, nil, AlertInternalError
+		}
+	}
+	if cct != nil {
+		err = ch.Extensions.Add(cct)
+		if err != nil {
+			logf(logTypeHandshake, "[ClientStateStart] Error adding ClientCertType extension [%v]", err)
+			return nil, nil, AlertInternalError
+		}
+		err = ch.Extensions.Add(sct)
+		if err != nil {
+			logf(logTypeHandshake, "[ClientStateStart] Error adding ServerCertType extension [%v]", err)
 			return nil, nil, AlertInternalError
 		}
 	}
--- a/client-state-machine.go
+++ b/client-state-machine.go
@@ -1072,7 +1072,7 @@ func (state clientStateWaitFinished) Next(hr handshakeMessageReader) (HandshakeS
 					certList[i] = CertificateEntry{CertData: entry.Raw}
 				}
 			} else {
-				certData, err := marshalSigningKey(cert.PublicKey)
+				certData, err := marshalSigningKey(cert.PrivateKey.Public())
 				if err != nil {
 					logf(logTypeHandshake, "[ClientStateWaitFinished] Unable to marshal raw public key [%v]", err)
 					return nil, nil, AlertInternalError
--- a/client-state-machine.go
+++ b/client-state-machine.go
@@ -620,11 +620,15 @@ func (state clientStateWaitEE) Next(hr handshakeMessageReader) (HandshakeState, 
 
 	serverALPN := &ALPNExtension{}
 	serverEarlyData := &EarlyDataExtension{}
+	serverClientCertType := &ClientCertTypeExtension{HandshakeType: HandshakeTypeEncryptedExtensions}
+	serverServerCertType := &ServerCertTypeExtension{HandshakeType: HandshakeTypeEncryptedExtensions}
 
 	foundExts, err := ee.Extensions.Parse(
 		[]ExtensionBody{
 			serverALPN,
 			serverEarlyData,
+			serverClientCertType,
+			serverServerCertType,
 		})
 	if err != nil {
 		logf(logTypeHandshake, "[ClientStateWaitEE] Error decoding extensions: %v", err)
@@ -635,6 +639,14 @@ func (state clientStateWaitEE) Next(hr handshakeMessageReader) (HandshakeState, 
 
 	if foundExts[ExtensionTypeALPN] && len(serverALPN.Protocols) > 0 {
 		state.Params.NextProto = serverALPN.Protocols[0]
+	}
+
+	if foundExts[ExtensionTypeClientCertType] &&
+		foundExts[ExtensionTypeServerCertType] &&
+		state.Config.AllowRawPublicKeys {
+		foundClient := serverClientCertType.CertificateTypes[0] == CertificateTypeRawPublicKey
+		foundServer := serverServerCertType.CertificateTypes[0] == CertificateTypeRawPublicKey
+		state.Params.UsingRawPublicKeys = foundClient && foundServer
 	}
 
 	state.handshakeHash.Write(hm.Marshal())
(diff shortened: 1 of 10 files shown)


client-state-machine.go
```diff
@@ -113,7 +131,7 @@ func (state clientStateStart) Next(hr handshakeMessageReader) (HandshakeState, [
 		return nil, nil, AlertInternalError
 	}
 	for _, ext := range []ExtensionBody{&sv, &sni, &ks, &sg, &sa} {
-		err := ch.Extensions.Add(ext)
+		err = ch.Extensions.Add(ext)
```
ekr, line 134: Why is this change needed? You seem to shadow it later everywhere relevant.

client-state-machine.go
```diff
@@ -128,6 +146,18 @@ func (state clientStateStart) Next(hr handshakeMessageReader) (HandshakeState, [
 			return nil, nil, AlertInternalError
 		}
 	}
+	if cct != nil {
```
ekr, line 149: Why did you split up the extension construction? it seems like you could just have this at the end of the cct, sct construction block.

client-state-machine.go
```diff
@@ -590,11 +620,15 @@ func (state clientStateWaitEE) Next(hr handshakeMessageReader) (HandshakeState,
 
 	serverALPN := &ALPNExtension{}
 	serverEarlyData := &EarlyDataExtension{}
+	serverClientCertType := &ClientCertTypeExtension{HandshakeType: HandshakeTypeEncryptedExtensions}
+	serverServerCertType := &ServerCertTypeExtension{HandshakeType: HandshakeTypeEncryptedExtensions}
```
ekr, line 624: I would call this :"selected"

client-state-machine.go
```diff
@@ -607,6 +641,14 @@ func (state clientStateWaitEE) Next(hr handshakeMessageReader) (HandshakeState,
 		state.Params.NextProto = serverALPN.Protocols[0]
 	}
 
+	if foundExts[ExtensionTypeClientCertType] &&
+		foundExts[ExtensionTypeServerCertType] &&
+		state.Config.AllowRawPublicKeys {
+		foundClient := serverClientCertType.CertificateTypes[0] == CertificateTypeRawPublicKey
+		foundServer := serverServerCertType.CertificateTypes[0] == CertificateTypeRawPublicKey
+		state.Params.UsingRawPublicKeys = foundClient && foundServer
```
ekr, line 649: BUG: what happens if the server doesn't want client auth at all but wants to use a raw public key?

client-state-machine.go
```diff
@@ -818,44 +860,67 @@ func (state clientStateWaitCV) Next(hr handshakeMessageReader) (HandshakeState,
...
+			logf(logTypeHandshake, "[ClientStateWaitCV] Unable to parse client cert: %v", err)
+		}
+
+		serverPublicKey = eeCert.PublicKey
+	} else {
+		serverPublicKey, err = unmarshalSigningKey(certs[0])
+		if err != nil {
+			logf(logTypeHandshake, "[ClientStateWaitCV] Unable to parse raw public key: %v", err)
+		}
+	}
```
ekr, line 883: BUG: You don't fail the handshake here.

extensions.go
```diff
@@ -624,3 +624,99 @@ func (c CookieExtension) Marshal() ([]byte, error) {
...
+//         select(ClientOrServerExtension) {
+//             case client:
+//               CertificateType client_certificate_types<1..2^8-1>;
+//             case server:
+//               CertificateType client_certificate_type;
+//         }
+// } ClientCertTypeExtension;
+type ClientCertTypeExtension struct {
+	HandshakeType    HandshakeType     `tls:"none"`
+	CertificateTypes []CertificateType `tls:"head=1,min=1"`
```
ekr, line 638: I haven't checked, but won't this effectively cause an assert here if you set the settings to not allow any types.

extensions.go
```diff
@@ -624,3 +624,99 @@ func (c CookieExtension) Marshal() ([]byte, error) {
...
+}
+
+func (sct *ServerCertTypeExtension) Unmarshal(data []byte) (int, error) {
+	switch sct.HandshakeType {
+	case HandshakeTypeClientHello:
+		return syntax.Unmarshal(data, sct)
+
+	case HandshakeTypeEncryptedExtensions:
+		var val uint8
+		n, err := syntax.Unmarshal(data, &val)
```
ekr, line 712: Are you checking length, here and above?

server-state-machine.go
```diff
@@ -344,6 +349,31 @@ func (state serverStateStart) Next(hr handshakeMessageReader) (HandshakeState, [
...
+			foundServer := false
+			for _, t := range clientServerCertType.CertificateTypes {
+				if t == CertificateTypeRawPublicKey {
+					foundServer = true
+					break
+				}
+			}
+
+			// Only support symmetric usage of raw public keys for now
+			connParams.UsingRawPublicKeys = foundClient && foundServer
```
ekr, line 374: BUG?: You should be failing negotiation here if the client asks for RawPublicKeys and not X.509 and you don't allow RawPublicKeys.

server-state-machine.go
```diff
@@ -591,6 +621,27 @@ func (state serverStateNegotiated) Next(_ handshakeMessageReader) (HandshakeStat
 			return nil, nil, AlertInternalError
 		}
 	}
+	if state.Params.UsingRawPublicKeys {
+		logf(logTypeHandshake, "[server] sending ClientCertType extension")
```
ekr, line 625: BUG: You are violating a spec requirement here:
```   If the server does not send a certificate_request payload (for
   example, because client authentication happens at the application
   layer or no client authentication is required) or none of the
   certificates supported by the client (as indicated in the
   client_certificate_type extension in the client hello) match the
   server-supported certificate types, then the client_certificate_type
   payload in the server hello MUST be omitted.
```

server-state-machine.go
```diff
@@ -643,13 +694,24 @@ func (state serverStateNegotiated) Next(_ handshakeMessageReader) (HandshakeStat
 			handshakeHash.Write(crm.Marshal())
 		}
 
-		// Create and send Certificate, CertificateVerify
-		certificate := &CertificateBody{
-			CertificateList: make([]CertificateEntry, len(state.cert.Chain)),
-		}
-		for i, entry := range state.cert.Chain {
-			certificate.CertificateList[i] = CertificateEntry{CertData: entry}
+		// Create and send Certificate
```
ekr, line 697: I would factor this out into a function.

ekr (changes requested):
Note: this is RFC 7250, not 7520.

Tests seem a bit thin.

client-state-machine.go
```diff
@@ -102,6 +102,24 @@ func (state clientStateStart) Next(hr handshakeMessageReader) (HandshakeState, [
 		alpn = &ALPNExtension{Protocols: state.Opts.NextProtos}
 	}
 
+	// Raw public keys
+	var cct *ClientCertTypeExtension
+	var sct *ClientCertTypeExtension
```
upros, line 107: Should be:
var sct *ServerCertTypeExtension

client-state-machine.go
```diff
@@ -102,6 +102,24 @@ func (state clientStateStart) Next(hr handshakeMessageReader) (HandshakeState, [
...
+	if state.Config.AllowRawPublicKeys || !state.Config.ForbidCertificates {
+		cct := &ClientCertTypeExtension{HandshakeType: HandshakeTypeClientHello}
+		sct := &ServerCertTypeExtension{HandshakeType: HandshakeTypeClientHello}
```
upros, line 110: should be:
cct = &ClientCertTypeExtension{HandshakeType: HandshakeTypeClientHello}
sct = &ServerCertTypeExtension{HandshakeType: HandshakeTypeClientHello}

server-state-machine.go
```diff
@@ -122,6 +123,8 @@ func (state serverStateStart) Next(hr handshakeMessageReader) (HandshakeState, [
 	clientALPN := new(ALPNExtension)
 	clientPSKModes := new(PSKKeyExchangeModesExtension)
 	clientCookie := new(CookieExtension)
+	clientClientCertType := new(ClientCertTypeExtension)
+	clientServerCertType := new(ServerCertTypeExtension)
```
upros, line 127: Should be:

clientClientCertType := &ClientCertTypeExtension{HandshakeType: HandshakeTypeClientHello}
clientServerCertType := &ServerCertTypeExtension{HandshakeType: HandshakeTypeClientHello}

client-state-machine.go
```diff
@@ -102,6 +102,24 @@ func (state clientStateStart) Next(hr handshakeMessageReader) (HandshakeState, [
 		alpn = &ALPNExtension{Protocols: state.Opts.NextProtos}
 	}
 
+	// Raw public keys
+	var cct *ClientCertTypeExtension
+	var sct *ClientCertTypeExtension
+	if state.Config.AllowRawPublicKeys || !state.Config.ForbidCertificates {
```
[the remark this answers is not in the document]
bifurcation: I don't see how that would arise.  If we take the branch, then one of those two conditions is true.  If the former, then `CertificateTypeRawPublicKey` is added; if the latter, then `CertificateTypeX509` is added.  So either you have at least one entry or `cct == sct == nil`, and the logic below does not add them if they are nil.

client-state-machine.go
bifurcation: Reverted.

client-state-machine.go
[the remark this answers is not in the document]
ekr: Maybe I don't understand what Config.ForbidCertificates is supposed to do.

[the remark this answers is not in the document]
bifurcation: `Config.ForbidCertificates = true` means that `CertificateTypeX509` is not included in the certificate type extensions, if they are sent.  I agree that's confusing.  ISTM there are two plausible refactors here, both assuming removal of `ForbidCertificates`:

1. Assume that X.509 is allowed, so `AllowRawPublicKeys` means that the cert type extension will have both values.  This is simple, but constrains the options available to the application.
2. Just have explicit type arrays in the `Config` object.  This is more flexible, but we might have to check for inconsistent config.

client-state-machine.go
bifurcation: Parallelism with how ALPN works already.  Refactoring both.

client-state-machine.go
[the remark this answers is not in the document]
ekr: I think I'm missing something, because if !AllowedRawPublicKeys && ForbidCertificates, then isn't the handshake impossible? So aren't you sending the extension for any valid config?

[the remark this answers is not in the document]
bifurcation: True, it'll just fail later.  And obviously, in the default X.509-only case, we don't need to send the extensions at all.  So it seems like the correct logic here (without changing the config format) would be something like:

```
if state.Config.AllowRawPublicKeys {
    // create extensions with RPKs allowed
    if !state.Config.ForbidCertificates {
        // add X509 to extensions
    }
}
```

But of course that muddies the definition of ForbidCertificates. Maybe the right refactor here is `ForbidCertificates -> RequireRawPublicKeys`, and:


```
if state.Config.AllowRawPublicKeys || state.Config.RequireRawPublicKeys {
    // create extensions with RPKs allowed
    if !state.Config.RequireRawPublicKeys {
        // add X509 to extensions
    }
}
```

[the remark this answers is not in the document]
ekr: I think I prefer the first, but I think either would work.

extensions.go
bifurcation: Correct, but that's what the RFC specifies.

client-state-machine.go
bifurcation: Good point.  Same actually applies for the certificate parsing a few lines above.

extensions.go
bifurcation: Yes.  `syntax.Unmarshal` will return an error if there is less than the required 1 byte available.  The fact that the entire extension body was consumed is checked by the `safeUnmarshal` call in `ExtensionList.Parse`.

extensions.go
ekr: Right I'm saying you might want this earlier up.

client-state-machine.go
bifurcation: I ended up refactoring the server logic here because of the bug you noted below.

bifurcation:
@ekr This update is substantially cleaned up, should address your comments.

PR #210, opened 2019-12-13
Add support for optional<T>

bifurcation:
Required for implementing MLS structs.  A pointer field in a struct can be tagged as "optional".  The null optional ("0x00") corresponds to the nil pointer.  Non-nil pointers are marshaled / unmarshaled as usual, with a 0x01 octet prefixed.

### DIFF
--- a/syntax/decode.go
+++ b/syntax/decode.go
@@ -28,10 +28,11 @@ type Unmarshaler interface {
 // These are the options that can be specified in the struct tag.  Right now,
 // all of them apply to variable-length vectors and nothing else
 type decOpts struct {
-	head   uint // length of length in bytes
-	min    uint // minimum size in bytes
-	max    uint // maximum size in bytes
-	varint bool // whether to decode as a varint
+	head     uint // length of length in bytes
+	min      uint // minimum size in bytes
+	max      uint // maximum size in bytes
+	varint   bool // whether to decode as a varint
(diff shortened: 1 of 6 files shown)


syntax/tags.go
```diff
@@ -27,6 +28,9 @@ func parseTag(tag string) tagOptions {
 		if token == varintOption {
 			opts[varintOption] = 1
 			continue
+		} else if token == optionalOption {
+			opts[optionalOption] = 1
+			continue
```
ekr: Should you forbid optional<T> on non-pointers?

syntax/decode.go
```diff
@@ -333,8 +335,26 @@ type pointerDecoder struct {
...
+		case 0:
+			indir := v.Elem()
+			indir.Set(reflect.Zero(indir.Type()))
+			return 1
+
+		case 1:
+			// No action; continue as normal
+
+		default:
+			panic(fmt.Errorf("Invalid flag byte for optional: [%x]", flag))
```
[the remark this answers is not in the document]
bifurcation: Yes.  This pattern was copied from `encoding/json`.

syntax/tags.go
bifurcation: This follows the same pattern as the other tags, and only interprets the tag when it's meaningful.  I'll do a separate "strictness" PR that checks alignment between tags and types.

PR #211, opened 2019-12-13
Enforce correct usage of struct tags

bifurcation:


syntax/tags_test.go
```diff
@@ -13,3 +14,37 @@ func TestTagParsing(t *testing.T) {
...
+		t.Fatalf("Accepted invalid tags for slice: %v", uintTags)
+	}
+
+	if tagsValidForType(ptrTags, uintType) {
+		t.Fatalf("Accepted invalid tags for uint: %v", ptrTags)
+	}
+
+	if tagsValidForType(sliceTags, ptrType) {
+		t.Fatalf("Accepted invalid tags for ptr: %v", sliceTags)
+	}
```
ekr: Don't you need 9 tests here?

bifurcation: Even 9 tests wouldn't be sufficient, since there are other types :)  It seems enough to verify failure in at least one case.

